PR #27, opened 2023-07-24
workfows: Add release.yml

jku:
Three steps:
* build signer
* Make signer release to pypi
* Make overall release on GitHub

This means at the moment signer and repository releases are tied to each other.

The release environment has been created: deployment requires approval from one of [jku, kommendorkapten, lukpueh].

The PyPI Trusted Publishing setup is "pending": it should become an actual setup on first release.

I have tested this (except for the pypi publishing part: that can only succeed on the upstream project)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,87 @@
+name: Release
+concurrency: release
+
+on:
+  push:
+    tags:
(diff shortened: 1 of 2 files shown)


kommendorkapten (approved):
Awesome work!

PR #28, opened 2023-07-24
Fix issue #26

### LINKED ISSUE #26: actions: yaml "linting"
yaml does wild things in its implicit value typing: let's try to avoid that by always using "" when values are strings: this is relevant at least for version numbers. As an example `python-version: 3.10` is incorrect and will lead to issues, should be `python-version: "3.10"`

For actions `.github/workflows/*.yml`, `actions/*/action.yml`:
1. make sure yaml string values, or at least version numbers, are in quotes
1. standardize to either single or double quotes by default
1. as a bonus we could setup some yaml linter in CI (assuming a good linter with reasonable defaults exists)

These do not have to happen all at once: any single one is an improvement

romw314:
Added quotes to every `actions/*/action.yml`.

resolved #26

### DIFF
--- a/actions/offline-version-bump/action.yml
+++ b/actions/offline-version-bump/action.yml
@@ -10,7 +10,7 @@ runs:
 
     - uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1
       with:
-        python-version: 3.11
+        python-version: "3.11"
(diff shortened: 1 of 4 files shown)


jku (approved):
Thanks, looks good.

PR #33, opened 2023-07-26
signer: Add import tool

jku:
@kommendorkapten not sure if this is too late for you but... `tuf-on-ci-import-repo <EVENT> [<IMPORTDATA>]` will import
* legacy playground repos
* generic tuf repositories (when given additional import configuration file which the tool can pre-generate)

This is a draft since 
* it's quite untested
* I'm unsure if adding something like this into the core code base is smart: it's just 200 lines but it's unlikely to get tested well as it will be only ran a few times
* needs documentation that explains the steps of importing
  * create a repository with the original metadata in the expected place
  * add workflows from tuf-on-ci-template
  * setup signer tool locally
  * handle unsupported keys somehow
  * run tuf-on-ci-import-repo to start a signing event

still, I think it's surprisingly self contained: I only needed to add one hack into the SignerRepository: otherwise it's a separate tool that cannot break anything else

### DIFF
--- a/repo/tuf_on_ci/_repository.py
+++ b/repo/tuf_on_ci/_repository.py
@@ -134,7 +134,13 @@ class CIRepository(Repository):
         keys = []
         for keyid in r.keyids:
             try:
-                keys.append(delegator.get_key(keyid))
+                key = delegator.get_key(keyid)
+                if known_good and "x-tuf-on-ci-keyowner" not in key.unrecognized_fields:
+                    # this is allowed for repo import case: we cannot identify known
+                    # good keys and have to trust that delegations have not changed
+                    continue
+
+                keys.append(key)
             except ValueError:
                 pass
         return keys
--- a/repo/tuf_on_ci/_repository.py
+++ b/repo/tuf_on_ci/_repository.py
@@ -539,3 +539,14 @@ class CIRepository(Repository):
             return True
 
         return False
+
+    def is_verified(self, rolename: str) -> bool:
+        if rolename in ["root", "timestamp", "snapshot", "targets"]:
+            delegator = self.open("root")
+        else:
+            delegator = self.open("targets")
+        try:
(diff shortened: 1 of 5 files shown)


kommendorkapten:
Did a quick scan, and it looks good. Just one thing that I ran into yesterday when I migrated/imported our repo from `playground-repository` to `tuf-on-ci`. 
Just importing it wasn't enough, as when I created a new root (signing event) with the updated metadata (`x-tuf-on-ci-*`) the signer bailed out, as when it was looking through the old (and signed `root.json`) it couldn't find if the user was a root key holder or not. I wasn't sure if this tool covered that scenario?

A possible solution to add backwards compatibility is to ignore the `x-tuf-on-ci-keyowner` and match on key-id instead.

jku:
> the signer bailed out, as when it was looking through the old (and signed `root.json`) it couldn't find if the user was a root key holder or not. I wasn't sure if this tool covered that scenario?

I handled this for the actual import command (this is the "one hack" I mentioned), but I think the same issue may come up for whoever needs to sign -- not sure yet as I haven't tried this fully through

kommendorkapten:
> I handled this by skipping the check for keys in "known good" metadata for the actual import command (this is the "one hack" I mentioned

Ah yes, makes sense. Re-read the code now and get it.

As of having this in `main`, I'm mostly positive as it lowers the barrier for adoption.

jku:
It's not perfect (and is not documented in the main README by design) but I have used this to create https://github.com/jku/rootsigning-staging as a test. I think we could merge this with the expectation that future imports likely will need some workarounds or tweaks to this code: Importing is **not** part of the normal process like the other commands

kommendorkapten (approved):
Lets merge as-is. As @jku said, import is most likely not a common process, so even if the tool is a bit cumbersome and may require som custom modification, the current state is good enough for a starting ground 👍

PR #34, opened 2023-07-28
actions: Use repos API instead of search API

jku:
(PR moved from playground)

Former is arguably the correct choice (search API is global, not project specific) and we have seen at least one case where the search API did not have the content that was already available in repos API and this lasted for a few hours.

Differences:
* new API lists only open issues by default
* new API does not allow filtering by issue type (so we theoretically could find PRs too): I don't think this is an issue as the PRs do not get a label automatically and we are searching based on label

### DIFF
--- a/actions/signing-event/action.yml
+++ b/actions/signing-event/action.yml
@@ -72,12 +72,14 @@ runs:
 
           issue = 0
           const repo = context.repo.owner + "/" + context.repo.repo
-          const issues = await github.rest.search.issuesAndPullRequests({
-            q: "label:" + process.env.GITHUB_REF_NAME + "+state:open+type:issue+repo:" + repo,
+          const issues = await github.rest.issues.listForRepo({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            labels: [process.env.GITHUB_REF_NAME],
           })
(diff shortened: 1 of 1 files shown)


jku:
I'm using this same technique in #176 so if that goes in we should maybe merge this one too

actions/signing-event/action.yml
```diff
@@ -35,12 +35,14 @@ runs:
 
           issue = 0
           const repo = context.repo.owner + "/" + context.repo.repo
-          const issues = await github.rest.search.issuesAndPullRequests({
-            q: "label:" + process.env.GITHUB_REF_NAME + "+state:open+type:issue+repo:" + repo,
+          const issues = await github.rest.issues.listForRepo({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            labels: process.env.GITHUB_REF_NAME,
```
jku, line 41: I believe this works but using an array here would be more correct I think

```suggestion
            labels: [process.env.GITHUB_REF_NAME],
```

jku:
rebased and fixed the labels argument to be an array (a string is also documented to work but this seems more correct)

kommendorkapten (approved):

PR #45, opened 2023-08-08
signer: Refactor user configuration

jku:
## Goals

* better code structure
* ability to configure signing systems in `.tuf-on-ci-sign.ini`

## Contents

Refactoring:
* Separate a _user module with a User class
* New class manages the users signers and configuration
* Design would allow User class to write the config (it does not currently)

New feature:
* privkeyuris can now be included in the configuration. These uris will then be used when signing. Example:

```
    [signing-keys]
    762cb22caca65de5e9b7b6baecb84ca989d337280ce6914b6440aea95769ad93 = hsm:2?label=YubiKey+PIV+%2315835999
```
The obvious next steps would be to automatically write the privkeyuris to the config when user becomes a signer. Technically this is possible to add already but I'm trying to keep the PR limited...

### DIFF
--- a/signer/test/test_user.py
+++ b/signer/test/test_user.py
@@ -1,0 +1,112 @@
+import os
+import unittest
+from tempfile import TemporaryDirectory
+
+import click
+from securesystemslib.signer import HSMSigner, SSlibKey
+
+from tuf_on_ci_sign._user import User
+
+# Long lines are ok here
+# ruff: noqa: E501
+
+REQUIRED = """
+[settings]
+pykcs11lib = /usr/lib/x86_64-linux-gnu/libykcs11.so
+user-name = @signer
+push-remote = origin
+pull-remote = myremote
+"""
+
+MISSING_NAME = """
+[settings]
+pykcs11lib = /usr/lib/x86_64-linux-gnu/libykcs11.so
+push-remote = origin
+pull-remote = myremote
+"""
+
+REQUIRED_AND_SIGNING_KEYS = """
+[settings]
+pykcs11lib = /usr/lib/x86_64-linux-gnu/libykcs11.so
+user-name = @signer
+push-remote = origin
+pull-remote = myremote
+
+[signing-keys]
+762cb22caca65de5e9b7b6baecb84ca989d337280ce6914b6440aea95769ad93 = hsm:2?label=YubiKey+PIV+%2315835999
+01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b = file:keys/mykey?encrypted=false
+"""
+
+HSM_KEY = SSlibKey(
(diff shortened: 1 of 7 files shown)


signer/tuf_on_ci_sign/_user.py
```diff
@@ -0,0 +1,75 @@
...
+    return click.style(text, bold=True)
+
+
+class User:
+    """Class that manages user configuration and manages the users signer cache"""
+
+    def __init__(self, path: str):
+        self._config_path = path
+
+        self._config = ConfigParser(interpolation=None)
```
jku, line 18: the argument is added because
* we don't need the fancy interpolation feature of ConfigParser
* interpolation prevents using some characters in config values and we want to use them in priv key uris

signer/test/test_user.py
```diff
@@ -0,0 +1,112 @@
...
+REQUIRED_AND_SIGNING_KEYS = """
+[settings]
+pykcs11lib = /usr/lib/x86_64-linux-gnu/libykcs11.so
+user-name = @signer
+push-remote = origin
+pull-remote = myremote
+
+[signing-keys]
+762cb22caca65de5e9b7b6baecb84ca989d337280ce6914b6440aea95769ad93 = hsm:2?label=YubiKey+PIV+%2315835999
+01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b = file:keys/mykey?encrypted=false
```
lukpueh, line 37: Is the file signer URI ignored in the tests below on purpose?

signer/test/test_user.py
```diff
@@ -0,0 +1,112 @@
...
+            self.assertIsInstance(other_signer, HSMSigner)
+            self.assertEqual(other_signer.token_filter, {})
+            self.assertEqual(
+                other_signer.public_key.keyid,
+                "64eeece964e09c058ef8f9805daca546b01ba4719c80b6fe911b091a7c05124b",
+            )
+
+            # another lookup should return same instance
+            second_hsm_signer = user.get_signer(HSM_KEY)
+            self.assertIs(hsm_signer, second_hsm_signer)
```
lukpueh, line 108: Would be nice (and probably easy) to add a "sigstore-oidc" test case. Then you'd have all cases in the `get_signer` switch tested.

lukpueh (approved):
I think the refactor makes sense, the code looks good, and the new feature is definitely helpful.

signer/test/test_user.py
jku: ah the sigstoresigner can't be tested like this because we currently authenticate at Signer instantiation time: this is arguably a mistake, maybe it should only auth on first sign()

signer/test/test_user.py
jku: not really, same story as sigstore: I didn't think it was worth it (but wanted an example with more than one key)

I was planning to add it now but since sigstore is not possible I won't bother with this either

PR #77, opened 2023-09-13
Fixed a bug that causes incorrect management of keys.

kommendorkapten:
When a signer (or multiple) are being removed, the resulting list of signers will not always be correct, as the list being iterated over is mutated during the iteration.


jku (approved):
Thanks, good catch.

This is an annoying feature of the python-tuf API: everything is a mutable reference...

PR #80, opened 2023-09-22
Actions refactor

jku:
The goals are:
* Make actions and workflows simpler and more logical
* Remove duplication currently present in workflows and actions
* Separate publishing from online signing to let individual projects have 
  more control over their publishing pipeline

I think this turned out pretty good (with some unforeseen side effects, see "other notes" below).
The question is mostly, is this worth the new bugs and trouble for existing repositories? 
I would lean on yes, it's worth it but willing to discuss. 

See sibling workflow PR https://github.com/theupdateframework/tuf-on-ci-template/pull/6 and  test repository in https://github.com/jku/refactor-test

#### workflow: create-signing-events

* Runs on schedule
* creates signing event branches and issues for offline roles that need
  signing events.
* Uses action `create-signing-events`

#### workflow: signing-event

* Runs on 
  * push (to signing event branches)
  * workflow dispatch (from create-signing-events)
* Updates the state of the signing event in the signing event issue
* Uses action `signing-event`

#### workflow: online-sign

* Runs on
  * schedule
  * push to main
* Signs snapshot and timestamp if needed, moves the "publish" branch if needed
* Uses action `online-sign`

#### workflow: publish

* Runs on workflow dispatch (from online-sign)
* Builds publishable repo, deploys it to GH pages
* _This is the one workflow projects should modify to their own needs_
* Uses action `upload-repository` (and deploy-pages from GH)

## Other notes

* I think the design is much cleaner: there is now mostly one action for workflow and they are logically named.
* Testing is still a pain: my first version had many bugs
* The concept of "publish" branch (to document the current publishable commit) seems good, even if
  it requires using workflow dispatch to trigger the publish workflow: Now the publish workflow really is something
  individual projects can modify as they want
* upload-repository action (used in publish) has the default mode of producing separate metadata and targets artifacts:
  I believe this is the correct design... unfortunately GH pages upload is a bit clunky so I left a special gp-pages mode
  in there
* publish from anything but main branch requires a config change
* I've left a few comments that I'd like to still act on but I believe other review is now useful


repo/tuf_on_ci/_repository.py
```diff
@@ -540,13 +570,24 @@ def update_targets(self, rolename: str) -> bool:
 
         return False
 
-    def is_verified(self, rolename: str) -> bool:
+    def is_valid(self, rolename: str) -> bool:
```
jku, line 573: may have to rethink the name here:
* there is already a `_validate_role()` that is used during a signing event (so takes into account the signing event starting point)
* this one is more about "does this role need to be signed" and is currently only used during online sign

actions/online-sign/action.yml
```diff
@@ -0,0 +1,71 @@
...
+          echo "ONLINE_SIGNED=true" >> "$GITHUB_ENV"
+        fi
+      shell: bash
+
+    - id: move-publish-branch
+      if: inputs.force_publish == 'true' || env.ONLINE_SIGNED == 'true'
+      run: |
+        git show --oneline --no-patch HEAD
+        git push origin HEAD:publish
+      shell: bash
```
jku, line 58: moving the publish branch currently happens here, in the step following the `tuf-on-ci-online-sign` call.

The alternative would be to do it in `tuf-on-ci-online-sign` (this would make some sense as it already pushes to "main"). The only thing needed for that is forwarding `force_publish` as an argument.

I think this alternative makes more sense as then the git pushes are in a single place

repo/tuf_on_ci/online_sign.py
```diff
@@ -28,35 +28,35 @@ def _git(cmd: list[str]) -> subprocess.CompletedProcess:
...
-    snapshot_updated, _ = repo.do_snapshot(not verified)
+    valid = repo.is_valid("snapshot")
+    snapshot_updated, _ = repo.do_snapshot(not valid)
     if not snapshot_updated:
         click.echo("No snapshot needed")
     else:
-        verified = repo.is_verified("timestamp")
-        repo.do_timestamp(not verified)
+        valid = repo.is_valid("timestamp")
+        repo.do_timestamp(not valid)
```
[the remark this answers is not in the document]
jku: should be fixed now

actions/online-sign/action.yml
```diff
@@ -0,0 +1,71 @@
+name: 'Online sign'
+description: 'Creates a snapshot and timestamp if needed, moves publish branch if needed'
+
+inputs:
+  force_publish:
+    description: 'Publish even when snapshot and timestamp are not created'
+    required: true
```
jku, line 7: As mentioned in the tuf-on-ci-template comment: Maybe this is not needed and the action should figure this out: if `$GITHUB_EVENT_NAME` is not "schedule" we want force_publish.

actions/online-sign/action.yml
kommendorkapten: Without thinking too much, I prefer to have the push to publish separated into a different steps, it feels like we are having better separation of duties this way? The command always work against `main`, but the workflows are in charge of cordinate all the steps?

docs/CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
 # Changelog
 
+## Unreleased
+
+NOTE: This is a major API break, users should **not** just upgrade the action versions but
+repalce their workflows with newworkflows from tuf-on-ci-template.
```
kommendorkapten, line 6:
```suggestion
repalce their workflows with new workflows from tuf-on-ci-template.
```

actions/online-sign/action.yml
jku: Yeah that is the thinking that lead me to write it this way: that pushing to "publish" is a separate decision from pushing to "main"...

But after implementing it I'm not so sure: 
* there's no checks I can think of that would make sense to do at this point (that wouldn't also make sense before merging to main) and there is a separate publish workflow already.
* keeping functionality in the workflow (instead of the tool) makes it harder to test

So I think pushing to "publish" at the same time as pushing to "main" would make more sense... but this is not a strong opinion.

actions/online-sign/action.yml
jku: This is done: I will leave this unresolved until I've seen a scheduled workflow work correctly

actions/online-sign/action.yml
jku: I will resolve this without changes:
* Changing this is not directly necessary for this PR: this is now an implementation detail of `online-sign` action and we can improve this later
* I think the change makes sense... but it's not actually trivial: I would rather have it separately reviewed if it's done

docs/REPOSITORY-MAINTENANCE.md
```diff
@@ -6,7 +6,10 @@ ongoing maintenance.
 ## New Repository Setup
 
 1. Fork the [template](https://github.com/theupdateframework/tuf-on-ci-template).
-1. To enable repository publishing, set _Settings->Pages->Source_ to `Github Actions`
+1. To enable repository publishing to GiHub Pages:
```
jku, line 9:
```suggestion
1. To enable repository publishing to GitHub Pages:
```

kommendorkapten (approved):
Nice work!

PR #91, opened 2023-10-04
Update SIGNER-SETUP.md

rdimitrov:


docs/SIGNER-SETUP.md
```diff
@@ -21,6 +21,7 @@ TUF-on-CI also needs access to a PKCS#11 module.
    macOS users can install with
    ```shell
    $ brew install yubico-piv-tool
+   $ brew install swig
```
jku, line 24: Hmm, swig isn't needed by HW signing but the signing tool install on line 40 I think.

So I think this needs some refactor -- maybe modify the "Signing tool installation" instructions to have separate instructions for mac and linux?

rdimitrov: Ah, you're right. Applied your feedback 👍

jku (approved):
yeah let's go with that

PR #92, opened 2023-10-04
Revert change of artifats path to targets.

kommendorkapten:
When publishing, the targets was moved to /artifacts insteaf of /targets if not specified. As this is a breaking change we should revert that.

### DIFF
--- a/actions/upload-repository/action.yml
+++ b/actions/upload-repository/action.yml
@@ -13,7 +13,7 @@ inputs:
   artifacts_path:
     description: 'relative published artifact path (only useful with gh_pages)'
     required: false
-    default: "artifacts"
+    default: "targets"
 
 runs:
   using: "composite"


kommendorkapten:
Updated a repository locally with this branch and verified that published targets are now available again in the `/targets` folder.

jku:
Right, I read my commits and refreshed my memory. This was intentional but I then forgot to add this to the changelog. I don't feel super strongly about this but I think we should consider modifying changelog as a fix as well.

The logic was this:
* targets is a confusing name in TUF: artifacts is much better
* the PR and existing code already use the "artifacts" term elsewhere so makes sense here
* it'll be easier to do changes like this today than it will be tomorrow (and this applies every day)

The one argument against is this:
* the directory in the source tree is still called "targets" -- my idea was to fix this slower in a compatible way (so new default would be artifacts but if "targets" dir exists, that is used) but I admit that might never happen

---

All that said, I'm fine just taking this PR.

kommendorkapten:
As the directory on disk is `targets` I think the default published directory should be `targets` too (follow the same pattern as with `metadata`). Beyond that I do agree artifacts may be a better name as `targets` is a bit overloaded in this context. But as this is configurable, and we are about to make a breaking change I think now is the time to revisit such decisions.

jku (approved):
let's go with this: changing this default to artifacts when the source directory is still "targets" is just confusing

PR #94, opened 2023-10-05
repo: Fix tuf-on-ci-status when artifacts are added to delegates

### LINKED ISSUE #93: Issue with artifact modification
@dimitrov found an issue with adding artifacts:

https://github.com/jku/tuf-demo/issues/68

I'm investigating, I think there's a bug in the code refactor

jku:
Fixes #93.

It seems this was never tested:
* the artifact path partition was just broken
* subdirectories were not handled properly

I don't have test yet: I think I'll file a new issue for that. I've tested this manually by running "tuf-on-ci-status --no-push" in the branch of this signing event https://github.com/jku/tuf-demo/issues/68: result looks good to me

The signing tool _should_ work just fine as it does not really look at the artifacts itself (arguably it maybe should but currently it does not)

CC @rdimitrov


jku:
I added a test. It's as bad as the e2e tests all are but it does work as regression test for this issue.

tests/e2e.sh
```diff
@@ -357,6 +355,27 @@ signer_modify_targets()
     git push --quiet origin $EVENT
 }
 
+signer_add_delegated_target()
```
jku, line 358: The test file really should get a file wide rename where "target" is changed to "artifact" where applicable: current state is confusing. But I didn't want to rename just the new methods and kept them in line with existing methods

kommendorkapten:
Just so I understand the scope, this only affects delegated targets in a directory inside `targets/`?

jku:
> Just so I understand the scope, this only affects delegated targets in a directory inside `targets/`?

Correct: 
* this is used in tuf-on-ci-status only for two purposes:
  * decide which roles metadata should be updated (update hashes to match the artifacts in  `targets/` )
  * decide which roles to list on the status message
* tuf-on-ci-status collects a list of roles that have changed in this signing event by looking at three things
  * roles whose metadata has changed
  * roles that have artifact changes in `targets/` directory <-- this is what's broken for delegated roles
  * roles that have open invites


(It's possible that we should be building this list of changed roles by comparing the artifacts in targets/ to the metadata, but that currently only happens later when tuf-on-ci-status prints the change details. This was just the first implementation that I thought of)

kommendorkapten (approved):
Great, just wanted to make sure I understood the PR.

Issue #79, opened 2023-09-22
feature: Add GitHub task summaries


jku:
It would be useful to do the most useful logging into step summaries: This is markdown GH renders on the workflow run page and is pretty easy to produce:

    echo "### Online signing was skipped as unnecessary" >> $GITHUB_STEP_SUMMARY

kommendorkapten:
As part of this task I'll intend to fix this error which happens if there is concurrent updates to the even branch during the execution of the `signing-event.yml` workflow:

```
Signers can sign these changes by running tuf-on-ci-sign sign/<event-name>
Git output on error: To https://github.com/<org>/<repo>
! [rejected] sign/<event-name> -> sign/<event-name> (fetch first)
error: failed to push some refs to 'https://github.com/<org>/<repo>'
hint: Updates were rejected because the remote contains work that you do not
hint: have locally. This is usually caused by another repository pushing to
hint: the same ref. If you want to integrate the remote changes, use
hint: 'git pull' before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

My plan for fixing this is to catch the error and detect that there are updates on remote, and so there is nothing to be done, as any changes that would be done (e.g. update `targets.json` to reflect changes to the targets) may be wrong (remote can contain new changes) and as there are changes, a new action would be triggered that would reconcile the state.

cc: @jku

PR #96, opened 2023-10-06
Adds step summaries for the most important steps

kommendorkapten:
Closes https://github.com/theupdateframework/tuf-on-ci/issues/79

Also this PR tries do detect updates on remote when running `tuf-on-ci-status`. If there are updates on remote, it wont print out the complete error message from git, instead it prints out that there are updates on remote, and so will not try to push.

As there are updates on remote, those updates will trigger a new workflow run that exectues `tuf-on-ci-status`, so it's safe to just ignore this error.

### DIFF
--- a/actions/signing-event/action.yml
+++ b/actions/signing-event/action.yml
@@ -64,9 +64,9 @@ runs:
             pr_url.searchParams.set("body", "Signing event " + process.env.GITHUB_REF_NAME + " is successful and ready to merge.\n\nCloses #" + issue + ".")
             message += "### Signing event is successful\n\n"
             message += "Threshold of signatures has been reached. A [pull request](" + pr_url + ") can be opened."
-            summary = "### Signing event is successful"
+            summary = "Signing event is successful"
           } else {
-            summary = "### Signing event in progress"
+            summary = "Signing event in progress"
           }
 
           github.rest.issues.createComment({
--- a/actions/signing-event/action.yml
+++ b/actions/signing-event/action.yml
@@ -33,6 +33,7 @@ runs:
         script: |
           const fs = require('fs')
           message = fs.readFileSync('./status-output').toString()
+          summary = ''
 
           issue = 0
           const repo = context.repo.owner + "/" + context.repo.repo
@@ -63,6 +64,9 @@ runs:
             pr_url.searchParams.set("body", "Signing event " + process.env.GITHUB_REF_NAME + " is successful and ready to merge.\n\nCloses #" + issue + ".")
             message += "### Signing event is successful\n\n"
             message += "Threshold of signatures has been reached. A [pull request](" + pr_url + ") can be opened."
+            summary = "### Signing event is successful"
+          } else {
+            summary = "### Signing event in progress"
           }
 
           github.rest.issues.createComment({
@@ -71,3 +75,5 @@ runs:
             repo: context.repo.repo,
             body: message,
           })
+
+          await core.summary.addHeading(summary).write()
--- a/actions/signing-event/action.yml
+++ b/actions/signing-event/action.yml
@@ -18,12 +18,11 @@ runs:
       run: |
         if tuf-on-ci-status >> status-output;  then
           echo "status=success" >> $GITHUB_OUTPUT
-          cat status-output >> "$GITHUB_STEP_SUMMARY"
         else
           echo "status=failure" >> $GITHUB_OUTPUT
-          echo "Failed to get status of event" >> "$GITHUB_STEP_SUMMARY"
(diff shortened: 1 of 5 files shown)


actions/signing-event/action.yml
```diff
@@ -18,8 +18,10 @@ runs:
       run: |
         if tuf-on-ci-status >> status-output;  then
           echo "status=success" >> $GITHUB_OUTPUT
+          cat status-output >> "$GITHUB_STEP_SUMMARY"
         else
           echo "status=failure" >> $GITHUB_OUTPUT
+          echo "Failed to get status of event" >> "$GITHUB_STEP_SUMMARY"
```
jku, line 24: This is not correct I think, the status return code is a bit unintuitive:
* tuf-on-ci-status exit code 0 means the signing event is "complete" (ready to merge)
* tuf-on-ci-status other exit codes means the signing event is not complete

 (obviously the exit code is also non-zero for actual bugs: this is not a well designed part)
 
 My suggestion here might be to
 * unconditionally print the status-output (see two lines below)
 * maybe print the summary variable in the following file-issue task (I assume doing that in js is not hard)
 
 (or just file another issue to handle this case)

actions/upload-repository/action.yml
```diff
@@ -51,3 +51,12 @@ runs:
       if: inputs.gh_pages == 'true'
       with:
         path: build/
+
+    - id: status-summary
+      shell: bash
+      run: |
+        if [ "${{inputs.gh_pages }}" == "true" ]; then
+            echo "Repository deployed to GitHub Pages" >> $GITHUB_STEP_SUMMARY
```
jku, line 59: nitpick: not really deployed at this point, only the artifact has been uploaded

actions/upload-repository/action.yml
```diff
@@ -51,3 +51,12 @@ runs:
...
+        else
+            echo "Repoistory is uploaded to artifacts" >> $GITHUB_STEP_SUMMARY
```
jku, line 61:
```suggestion
            echo "Repository is uploaded to artifacts" >> $GITHUB_STEP_SUMMARY
```

repo/tuf_on_ci/status.py
```diff
@@ -206,6 +202,17 @@ def status(verbose: int, push: bool) -> None:
...
-        _git(["push", "origin", event_name])
+        try:
+            _git(["push", "origin", event_name])
+        except subprocess.CalledProcessError as e:
+            # Figure out if this is an error caused by remote being ahead
+            # of local branch
+            msg = "Updates were rejected because the remote contains work that you do"
+            found = e.stdout.find(msg)
+            if found:
+                print("Updates on remote, will not push")
```
jku, line 213: I'm trying to think how this works in the event and I _think_ it is handling the important cases: if there are actual metadata changes that we did not push, then `success==False` because those metadata are now unsigned... So we won't end up with singing-event saying the signing event is finished.

The message could be less about git internal and more about "There are new changes in signing event: skipping metadata update" or something but that's a nitpick

jku (changes requested):
Thanks. I have one change request and a few nitpicks.

repo/tuf_on_ci/status.py
kommendorkapten: > So we won't end up with singing-event saying the signing event is finished even though we skipped the push.

Yes, my understanding is that those changes may actually have been reverted by the change on remote, so the workflow that would be triggered by that change would reconcile the state and then push that.

actions/signing-event/action.yml
kommendorkapten: I think it's a bit tricky to add a step summary in the `file-issue`, I can take a look at that next week.

kommendorkapten: As it's a `uses: ` step. But I think it should be possible.

kommendorkapten: Marking this a draft and see if I can fix this.

jku: :+1: Also fine to leave this part out of the PR and just file a new issue about it

kommendorkapten:
@jku this is now ready for review, see https://github.com/kommendorkapten/tuf-repo-test-4/actions/runs/6468593063 for an example run.

jku (approved):
Looks good.

I think we'll have to be careful about not making the actions more complex, and probably we also want to improve the messages in the summaries... but both of these make sense once we've seen if these summaries are useful and what would make them more useful

kommendorkapten:
> but both of these make sense once we've seen if these summaries are useful and what would make them more useful

Agreed.

PR #98, opened 2023-10-09
Update Dependabot config so it finds actions

jku:
This was missing from the refactor PR.


jku:
Merging this, I'd like to get the actions updated

PR #115, opened 2023-10-25
Update tuf version

jku:
Upgrade doesn't strictly speaking require any changes but I wanted to use the new get_verification_result() to test it...

This turns out to be much more work than I expected (this only handles repo codebase, not signer): 
* VerificationResult containing keyids turned out to be less useful: I added a wrapper VerificationResultWithKeys but would like to change the upstream VerificationResult in the end
* I've modified close() to use VerificationResult as well: this had a unexpected side effect where the order of keys was not reproducible (see sort() calls in VerificationResultWithKeys)
* we can remove `open_prev()` now: the simpler `_known_good(self, rolename: str) -> Root | Targets | None` works in all cases
* refactored `_get_signing_status()` so it's smaller and calls separate methods for finding invites and getting verification results
* the `status()` method was refactored a bit but the main change is that VerificationResultWithKeys contains Keys: code needs to lookup the signer name in the keys 

cc @lukpueh let's have a look at this at some point

### DIFF
--- a/repo/tuf_on_ci/_repository.py
+++ b/repo/tuf_on_ci/_repository.py
@@ -21,6 +21,7 @@ from tuf.api.metadata import (
     Metadata,
     MetaFile,
     Root,
+    Signed,
     Snapshot,
     TargetFile,
     Targets,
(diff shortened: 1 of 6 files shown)


jku:
This should be rewritten after theupdateframework/python-tuf#2551 is released

repo/tuf_on_ci/signing_event.py
```diff
@@ -144,7 +149,7 @@ def _role_status(repo: CIRepository, role: str, event_name) -> bool:
     if status.message:
         click.echo(f"**Error**: {status.message}")
 
-    return role_is_valid and not status.invites
```
jku, line 147: SigningStatus makes sure it is not valid if there are invites so this is safe

jku:
FYI @lukpueh in case you are interested.

repo/tuf_on_ci/_repository.py
```diff
@@ -286,11 +282,6 @@ def _validate_role(
             if not md.signed.consistent_snapshot:
                 return False, "Consistent snapshot is not enabled"
 
-            # Specification: root version must be x+1, not just larger
-            if prev_md and prev_md.signed != md.signed:
-                if md.signed.version != prev_md.signed.version + 1:
-                    return False, f"Version {md.signed.version} is not valid for root"
-
```
jku, line 293: This check is handled by `Root.get_root_verification_result()`

jku:
This update is now possible as tuf 4.0 has been released. There is no rush though: the securesystemslib pinning in #243 handles the main reason to upgrade quickly

jku:
This is still in limbo as tuf, sigstore and securesystemslib all have had changes that need releases before the three are compatible again :( Basically:
* we'd like to use tuf 4.0 for the new API
* but that's not possible before a new sigstore release (because of secusystemslib dependency incompatibility)

jku:
Closing this for now, we have enough open PRs without drafts

PR #119, opened 2023-11-03
Release v0.2.0

jku:
Bug fix release. signer actually has no changes but the release process currently can't handle not releasing signer as well so both get a version bump.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## v0.2.0
+
+* GitHub actions now output step summaries: these are visible in workflow
+  run pages on Github (#96)
+* Improved output in signing event status comments (#101)
+* Fixed online signing with ambient Sigstore identity, which broke in 0.1.0
+  (#118)
+
+Upgrade instructions from v0.1.0:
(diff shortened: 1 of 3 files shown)


lukpueh (approved):
Cross-checked changes in git with changelog. Not mentioned changes are
- #98
- #100
- any dependabot bumps

Seems fair to leave them out.

docs/CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## v0.2.0
+
+* GitHub actions now output step summaries: these are visible in workflow
+  run pages on Github (#96)
```
kommendorkapten, line 6: Should we make the references real links?

kommendorkapten (approved):

docs/CHANGELOG.md
jku: I think the Github release (and so the dependabot message) should look fine as is. Maybe you're still right though, will consider next time.

kommendorkapten: Yeah, "naked" links like #96 only works in the UI, in markdown docs they need to be "real" links like `[#96](...issues/96)`

Issue #56, opened 2023-08-15
Make KMS logins work the same way in snapshot / online-version-bump


jku:
Currently GCP KMS and Azure Key Vault are supported differently
* If user wants to use GCP, they 
  * define some variables in GitHub settings
  * select GCP when configuring the repo with tuf-on-ci-delegate
* If user wants to use Azure, they 
  * define some variables/secrets in GitHub settings
  * select Azure when configuring the repo with tuf-on-ci-delegate 
  * **AND** modify snapshot and version-bump workflows to do azure authentication

I would like it if the workflow modification wasn't necessary. I'm not 100% sure embedding the authentication (like we do for GCP) is the correct choice but it would be nice if the solution  for both was similar

jku:
The upside of the GCP solution is that it requires no modification of the online signing workflows which is a plus: upgrading to new workflow versions is easier when they inevitably change.

PR #120, opened 2023-11-08
Add support for AWS KMS

### LINKED ISSUE #117: Support for AWS KMS
It would be great to support signing keys in AWS KMS. Looking at the existing code it looks like this is already supported in securesystemslib (https://github.com/secure-systems-lab/securesystemslib/blob/b17caf3176783ee29dcd7453a92ee3fcd4aa76d4/securesystemslib/signer/_aws_signer.py), so would just be a case of updating tuf-on-ci-delegate to be able to import with the AWS key ID.

Let me know if that sounds right and I'll be happy to take a stab at the implementation!

jonnystoten:
Fixes #117.

Adds support for AWS KMS, which was basically already there due to the AWSSigner class from securesystemslib.

This PR adds support to tuf-on-ci-delegate for setting up online signing to use AWS KMS. As part of this we need to pass the scheme for the key, see https://github.com/secure-systems-lab/securesystemslib/issues/668. I've made this a multiple choice, displaying the schemes as AWS names them.

I also added an AWS auth step to the online signing action, just like the Google one. I saw there was some discussion on whether this is right way to go in #56, I think it's fine but welcome other thoughts. I'll be happy to look at switching Azure over to match too.

Tested over at https://github.com/jonnystoten/tuf-testing

### DIFF
--- a/actions/online-sign/action.yml
+++ b/actions/online-sign/action.yml
@@ -36,7 +36,7 @@ runs:
 
     - name: Authenticate to AWS
       if: inputs.aws_role_to_assume != ''
-      uses: aws-actions/configure-aws-credentials@v4
+      uses: aws-actions/configure-aws-credentials@010d0da01d0b5a38af31e9c3470dbfdabdecca3a #v4.0.1
       with:
         aws-region: ${{ inputs.aws_region }}
         role-to-assume: ${{ inputs.aws_role_to_assume }}
--- a/actions/online-sign/action.yml
+++ b/actions/online-sign/action.yml
@@ -1,15 +1,23 @@
-name: 'Online sign'
-description: 'Creates a snapshot and timestamp if needed, moves publish branch if needed'
+name: "Online sign"
+description: "Creates a snapshot and timestamp if needed, moves publish branch if needed"
 
 inputs:
   gcp_workload_identity_provider:
-    description: 'Google Cloud workload identity provider'
+    description: "Google Cloud workload identity provider"
     required: false
-    default: ''
+    default: ""
   gcp_service_account:
-    description: 'Google Cloud service account name'
+    description: "Google Cloud service account name"
     required: false
-    default: ''
+    default: ""
+  aws_region:
+    description: "AWS region"
+    required: false
+    default: ""
+  aws_role_to_assume:
+    description: "AWS role to assume"
+    required: false
+    default: ""
 
 runs:
   using: "composite"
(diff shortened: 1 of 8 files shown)


jku:
Did not get to this one yet, sorry: it's on my list for next week

jku (changes requested):
Looks good to me with some tweaks:
* linter has a whitespace complaint (you can run all linters locally with `tox -m lint`)
* I think changing the test key UI choice value from 4 to 5  is going to require a few additional changes:
  *  range for click.prompt() is currently (1,4)
  * end-to-end test `tests/e2e.sh` uses the value 4 in three places to choose the test key (search for "Enter online key type"

The end-to-end tests are very annoying to debug so feel free to ping me if you have issues: I would expect things to pass once you update key value mentioned above. Basics of testing are:
* `tox -m test` runs all tests
* the e2e tests can be run manually and with debug output (in the tests/ dir):  `DEBUG_TESTS=1 ./e2e.sh`
* With DEBUG_TESTS the working temp dir will be left in place so it's possible to see what went wrong... but it's admittedly a complicated setup and a hack

actions/online-sign/action.yml
```diff
@@ -26,6 +34,13 @@ runs:
         workload_identity_provider: ${{ inputs.gcp_workload_identity_provider }}
         service_account: ${{ inputs.gcp_service_account }}
 
+    - name: Authenticate to AWS
+      if: inputs.aws_role_to_assume != ''
+      uses: aws-actions/configure-aws-credentials@v4
```
jku, line 39: could use the last release commit hash + version comment here like the other actions do

jonnystoten:
Thanks for the review!

I've changed the testing key to option `0` so that we have room to expand in the future, and fixed tests.

I added a `git_clone` function in the e2e tests which always sets user/email config and also disables GPG signing. I was having issues because my global git config enables GPG signing.

I also had an issue with libfaketime - I needed to change the path to `/usr/lib/faketime/libfaketime.so.1` (from `/usr/lib/x86_64-linux-gnu/faketime/libfaketime.so.1`). I left this uncommitted because I guess it needs to be in the existing location for GHA and it can vary by distro, not sure what the best solution there is.

jku:
> I also had an issue with libfaketime - I needed to change the path to `/usr/lib/faketime/libfaketime.so.1` (from `/usr/lib/x86_64-linux-gnu/faketime/libfaketime.so.1`). I left this uncommitted because I guess it needs to be in the existing location for GHA and it can vary by distro, not sure what the best solution there is.

One option is to use the faketime binary instead... I cannot remember if there was a reason I didn't originally.

In any case, that doesn't need to be part of this PR. Thanks for persisting with the hacky tests.

jku (approved):
I think this looks great.

Hopefully we can remove _collect_key_scheme() later when securesystemslib improves but even if not, that's manageable. 

Thanks for improving e2e.sh -- every little bit helps there, it's a real hack.

kommendorkapten (approved):
Nice!

Issue #126, opened 2023-11-15
enhancement: Make sure correct version is published


kommendorkapten:
This has only been seen once.

What happened?
1. During online signing, the (possibly updated) branch is [pushed](https://github.com/theupdateframework/tuf-on-ci/blob/main/actions/online-sign/action.yml#L55) to `publish`
2. Once pushed, the [publish.yml](https://github.com/theupdateframework/tuf-on-ci/blob/main/actions/online-sign/action.yml#L65) workflow is dispatched (using `ref: publish`)
3. The [upload-action](https://github.com/theupdateframework/tuf-on-ci/blob/main/actions/upload-repository/action.yml#L21) workflow then checks out the ref (i.e `publish`) and builds the TUF repository to be published.

What's occurred once is that due to a delay in GitHub, the publish job checked out the previous revision! This lead to an same version being published, i.e not the updated one.

While this seems to be an unlikely event, we should protect against it.

Proposed solution:
When the online signing action pushes `HEAD` to `publish`, it should record the git commit, and provide that as an input to the `publish` workflow, which can then be used during checkout as an input parameter (i.e specify which commit to clone). If the commit is not visible due to a delay, the workflow would fail, which is visible and can be monitored, and so, an action can be taken.

Another option is to add some logic to verify we are publishing the expected commit, and if the local commit is not the expected, the job could retry a few times with some back-off time in between. 

For simplicity, and the fact that this seem to be a rare error state, I opt for just extracting the git commit and try to checkout that specific commit.

jku:
> When the online signing action pushes HEAD to publish, it should record the git commit, and provide that as an input to the publish workflow, which can then be used during checkout as an input parameter

Sounds good to me.

PR #127, opened 2023-11-15
Verify that correct version is published.

kommendorkapten:
Closes https://github.com/theupdateframework/tuf-on-ci/issues/126

This is 1/2 PRs, the other is against the template repo as it contains an update to the workflow.

During the online sign action, the produced commit is saved, and passed as an input parameter to the deploy workflow. The deploy workflow uses this ref when checking out the repository. This ref can also be set during manual invocation of the publish workflow. If nothing is provided the mos recent commit is used (the default behaviour in the checkout action).
If the commit from the signing action is not reachable (i.e due to a delay in GitHub) the checkout action fails, and abort the deployment. This is an actionable event and the action can so be re-run.

An example signing: https://github.com/kommendorkapten/tuf-repo-test-2/actions/runs/6877640752/job/18705952685#step:2:383

and the corresponding publish:
https://github.com/kommendorkapten/tuf-repo-test-2/actions/runs/6877751166/job/18705966971#step:2:10

### DIFF
--- a/actions/online-sign/action.yml
+++ b/actions/online-sign/action.yml
@@ -69,6 +69,6 @@ runs:
             workflow_id: 'publish.yml',
             ref: 'publish',
             inputs: {
-              ref: '${{ steps.move-publish-branch.outputs.rev }}',
+              ref: 'a'+'${{ steps.move-publish-branch.outputs.rev }}',
             },
           })
--- a/actions/online-sign/action.yml
+++ b/actions/online-sign/action.yml
@@ -69,6 +69,6 @@ runs:
             workflow_id: 'publish.yml',
             ref: 'publish',
             inputs: {
-              ref: 'a'+'${{ steps.move-publish-branch.outputs.rev }}',
+              ref: '${{ steps.move-publish-branch.outputs.rev }}',
             },
           })
--- a/actions/online-sign/action.yml
+++ b/actions/online-sign/action.yml
@@ -69,6 +69,6 @@ runs:
             workflow_id: 'publish.yml',
             ref: 'publish',
             inputs: {
-              ref: ${{ steps.move-publish-branch.outputs.rev }},
+              ref: '${{ steps.move-publish-branch.outputs.rev }}',
             },
           })
--- a/actions/online-sign/action.yml
+++ b/actions/online-sign/action.yml
@@ -53,6 +53,7 @@ runs:
       run: |
         git show --oneline --no-patch HEAD
         git push origin HEAD:publish
+        echo "rev=`git rev-parse HEAD`" >> $GITHUB_OUTPUT
         echo "### Online signing finished, will now publish" >> $GITHUB_STEP_SUMMARY
       shell: bash
 
@@ -67,4 +68,7 @@ runs:
             repo: context.repo.repo,
             workflow_id: 'publish.yml',
             ref: 'publish',
+            inputs: {
+              ref: ${{ steps.move-publish-branch.outputs.rev }},
+            },
           })
--- a/actions/upload-repository/action.yml
+++ b/actions/upload-repository/action.yml
@@ -14,11 +14,17 @@ inputs:
     description: 'relative published artifact path (only useful with gh_pages)'
     required: false
     default: "targets"
+  ref:
+    description: 'Ref to clone'
+    required: false
+    default: ''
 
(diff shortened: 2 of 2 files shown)


kommendorkapten:
Remainder that when this is released, it must document that the workflows must be updated too

actions/online-sign/action.yml
```diff
@@ -67,4 +68,7 @@ runs:
             repo: context.repo.repo,
             workflow_id: 'publish.yml',
             ref: 'publish',
+            inputs: {
+              ref: '${{ steps.move-publish-branch.outputs.rev }}',
+            },
```
jku, line 73: if we give the actual ref as argument to createWorkflowDispatch (instead of using "publish") would that work without the added action args? Or am I missing something?

jku: I may have to read the whole publishing code again... I'm not quite sure how it chooses the correct commit right now :joy:

kommendorkapten: Yeah, it can work! I have to look at how the official checkout action works! Good idea.

kommendorkapten: According to the doc, I don't think it would work. The `ref` has to be a branch name of tag: 
> The git reference for the workflow. The reference can be a branch or tag name.

kommendorkapten: And it kind of makes sense, as the action still makes a checkout, and so would use the tip of the provided branch.

jku: so my understanding/assumption:
* actions/checkout defaults to the ref of the event that triggers the workflow (I assume this holds even though the checkout happens as part of a composite action)
* so as long as the publish workflow is triggered with the ref we want, this all should work without additional inputs
* I assume we can just use hash ref as the ref argument to createWorkflowDispatch -- I don't think the docs are making the case that only branch or tag would work?

This should be simple to test...

kommendorkapten: > (I assume this holds even though the checkout happens as part of a composite action) 

Yes, if `ref` is empty (i.e not set which is the case today) it defaults to the ref that triggered the event, i.e `publish`.

> so as long as the publish workflow is triggered with the ref we want, this all should work without additional inputs

Yes, according to the doc, only branch or tag is allowed. but let me test that, I have a repo ready for that.

kommendorkapten: Also tried with building a ref from the commit:

>  message: 'No ref found for: git/commits/8a323a9f7132350da18c54f87f8623f7c021d680',

kommendorkapten: So the documentation seems to be correct. When dispatching a workflow, you need to provide a tag or a branch name.

jku: got it, thanks for testing

actions/upload-repository/action.yml
```diff
@@ -14,11 +14,17 @@ inputs:
     description: 'relative published artifact path (only useful with gh_pages)'
     required: false
     default: "targets"
+  ref:
+    description: 'Ref to clone'
+    required: false
+    default: ''
```
jku, line 20: this is related to my question in tuf-on-ci-template:
* what does '' mean as checkout ref?
* should this actually be `required: true` since we never use it without a ref arg

jku (approved):
looks good. Left one question, feel free to pass if you don't think it's relevant.

Let's have a chat on slack WRT merging and releasing -- maybe there's a good common way to handle incompatible tuf-on-ci / tuf-on-ci-template releases that we should start following.

actions/upload-repository/action.yml
kommendorkapten: The default is the empty string, which is forwarded to `actions/checkout`:
```
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''
```

So if this is not provided, it will work as it did before, i.e it pulls the "latest" commit. I wrote latest in "" as there is no guarantees that you will actually get the commit pushed in the signing job.

jku: pedantically I don't think the default ever gets used though: the empty value comes from the caller.

Anyway, if it works that's good enough for me :+1:

PR #147, opened 2023-12-08
Support custom GitHub tokens

jku:
## Add support for custom tokens

This PR adds support for a custom GitHub token. The idea is to keep the default GitHub token as a low-permission one (so other workflows don't operate on a high permission token unnecessarily) and to add a fine grained token with higher permissions into repository secrets.

Implementation requires some workarounds to support both default and custom tokens (see dispatching discussion below) and a bit of refactoring (see python command splitting). See docs/REPOSITORY-MAINTENANCE.md changes for documentation.

This is going to require users to modify their workflows (or rather copy the workflows from tuf-on-ci-template) on upgrade as the actions now require a token argument.

### repo code changes:
* `tuf-on-ci-status` command was previously provided by the `status` module. Refactor so that there is a `signing_event` module that provides both `tuf-on-ci-status` and `tuf-on-ci-update-targets` commands
*  `tuf-on-ci-status` is now simpler and does not modify metadata: it only outputs signing event status and sets return value based on that
*  `tuf-on-ci-update-targets` modifies targets metadata to match actual artifacts (in the same way -status used to do). If no changes are required it returns 1. If metadata changes are required, it commits changes, pushes them, outputs  a message describing what was done and returns 0

### action changes
* signing-event, online-sign and create-signing-events now _require_ a token argument
* dispatch mechanisms are tweaked to allow use of custom tokens: if the token used to make a "git push" is the default token, Github will **not** trigger a push event but if a custom token does the same, an event will be triggered. This means we need to handle the two cases differently: typically by manual dispatch that is only run if token is default token 
* _signing-event_ action now either does targets metadata updates or outputs event status: 
  * first it runs `tuf-on-ci-update-targets`. If changes were done, nothing else happens (except another signing-event workflow is manually dispatched for default token case). 
  * if no changes were done by update-targets, signing-event action then runs `tuf-on-ci-status` to output signing  event status

### workflow changes

[tuf-on-ci-template draft PR](https://github.com/theupdateframework/tuf-on-ci-template/pull/13)

* always provide a token to the actions using
```
        with:
          token: ${{ secrets.TUF_ON_CI_TOKEN || secrets.GITHUB_TOKEN }}
```
* allow users to define TUF_ON_CI_TOKEN secret. If they don't, GITHUB_TOKEN is used (so nothing changes by default)
* exclude the online role files from online-sign push trigger paths: this prevents annoying double triggers when online-sign pushes a commit that triggers another online-sign...

### DIFF
--- a/actions/create-signing-events/action.yml
+++ b/actions/create-signing-events/action.yml
@@ -35,6 +35,7 @@ runs:
       shell: bash
 
     - name: Dispatch signing event workflow
+      # dispatch if using default token: otherwise create-signing-events step has already triggered push events
       if: inputs.token == github.token && steps.create-signing-events.outputs.events != ''
       env:
         EVENTS: ${{ steps.create-signing-events.outputs.events }}
--- a/actions/create-signing-events/action.yml
+++ b/actions/create-signing-events/action.yml
@@ -1,11 +1,17 @@
 name: 'Create signing events'
 description: 'Create signing events for offline signed metadata that is about to expire'
 
+inputs:
+  token:
+    description: 'GitHub token'
+    required: true
+
 runs:
   using: "composite"
   steps:
     - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       with:
+        token: ${{ inputs.token }}
         fetch-depth: 0
 
     - uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c
@@ -34,6 +40,7 @@ runs:
         EVENTS: ${{ steps.create-signing-events.outputs.events }}
       uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea
       with:
+        github-token: ${{ inputs.token }}
         script: |
           console.log('Dispatching events: ', process.env.EVENTS)
           process.env.EVENTS.trim().split(' ').forEach(event => {
--- a/actions/create-signing-events/action.yml
+++ b/actions/create-signing-events/action.yml
@@ -35,7 +35,7 @@ runs:
       shell: bash
 
     - name: Dispatch signing event workflow
-      if: steps.create-signing-events.outputs.events != ''
+      if: inputs.token == github.token && steps.create-signing-events.outputs.events != ''
       env:
         EVENTS: ${{ steps.create-signing-events.outputs.events }}
       uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea
--- a/actions/signing-event/action.yml
+++ b/actions/signing-event/action.yml
@@ -1,6 +1,21 @@
 name: 'Signing event'
 description: 'TUF-on-CI Signing event management'
 
+# This action is called from signing-event workflow, which is dispatched in multiple ways
+# depending on the type of token provided in inputs.
+#
+# 1. create-signing-events action creates a new signing event branch
+#     * When using a custom token this triggers push event handler
+#     * When using the default GitHub token the action calls createWorkflowDispatch()
+# 2. A signer pushes artifact changes to a signing event branch
+#     * This triggers push event handler
+# 3. This action (signing-event) makes a metadata change in update_targets step as a result of an artifact change
+#     * When using a custom token this triggers push event handler
+#     * When using the default GitHub token the action calls createWorkflowDispatch()
+#
+# Cases 1 & 3 lead to status step running. Case 2 leads to skipping status step, but
+# triggering case 3 immediately afterwards.
+
 inputs:
   token:
     description: 'GitHub token'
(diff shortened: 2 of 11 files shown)


jku:
Example repository running this code:
https://github.com/jku/test-custom-token/

Example signing event that demonstrates the way artifact changes now get their own automatic comment:
https://github.com/jku/test-custom-token/issues/17

jku:
> ### workflow changes
> [tuf-on-ci-template draft PR](https://github.com/theupdateframework/tuf-on-ci-template/pull/13)
> 
> * always provide a token to the actions using
> 
> ```
>         with:
>           token: ${{ secrets.TUF_ON_CI_TOKEN || secrets.GITHUB_TOKEN }}
> ```
> 
> * allow users to define TUF_ON_CI_TOKEN secret. If they don't, GITHUB_TOKEN is used (so nothing changes by default)
> * exclude the online role files from online-sign push trigger paths: this prevents annoying double triggers when online-sign pushes a commit that triggers another online-sign...

A fair review comment on the workflows is that when using TUF_ON_CI_TOKEN, the `permissions` item in the workflow could now be restricted... but that would mean two different sets of workflows (one set for custom token, others for default token) which I'm not fond of.

actions/signing-event/action.yml
```diff
@@ -77,3 +112,17 @@ runs:
           })
 
           await core.summary.addHeading(summary).write()
+
+    - name: Dispatch another signing event workflow
```
haydentherapper, line 116: Just curiosity, what is the reason the workflow is kicked off again?

jku: The issue is different behaviour for custom and default token: since the workflow is triggered by push, the workflow may automatically "self-trigger" (if it does make a new commit for the targets metadata changes) when using a custom token _but not when using the default token_... The options that I see for solving this are:
* Avoid push triggers completely somehow (I don't think this is practical)
* Let the signing-event sometimes run twice but exit early if we think we already ran once for the same commit
* Make sure the signing-event always runs twice after targets metadata changes

I've chosen the 3rd option because it actually fits the process. There are now two possible paths when the signing event is triggered:
1. targets metadata changes are needed: only do those changes, then let signing-event trigger again (either automatically or by manual dispatch)
1. targets metadata changes are not needed: output signing event status

jku: I tried to document this at the top of the action but as you can see from the previous comment it's I've not exactly figured how to explain the logic succinctly...

haydentherapper: One part that is unclear to me is:

```
#     * When using a custom token this triggers push event handler
#     * When using the default GitHub token the action calls createWorkflowDispatch()
```

Why is there a different action taken with a custom token vs the default token? I would have thought that by providing a custom token, you're simply overriding the default and the same path is taken.

jku: It's how GitHub works:

> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run.

That could be more explicit, what it's not saying out loud is that _events triggered by other tokens will create new workflow runs_.

haydentherapper: Ah, thanks, didn’t realize that restriction exists. I guess it’s to prevent using up available quota? Nonetheless, makes sense why this is set up this way.

actions/create-signing-events/action.yml
```diff
@@ -29,11 +35,12 @@ runs:
       shell: bash
 
     - name: Dispatch signing event workflow
-      if: steps.create-signing-events.outputs.events != ''
+      if: inputs.token == github.token && steps.create-signing-events.outputs.events != ''
```
kommendorkapten, line 38: Here I think a comment that `GITHUB_TOKEN` does not automatically dispatch a workflow would help. It's documented in `actions/signing-event/action.yml` but that context may be missing.

docs/REPOSITORY-MAINTENANCE.md
```diff
@@ -87,3 +87,32 @@ use.
...
+
+tuf-on-ci workflows (with the exception of `publish`) are written in a way to minimize
+need to modify the workflows: It may be useful to consider the workflows part of the
+tuf-on-ci application. The intention with this is to make workflow upgrades easier:
+tuf-on-ci release notes will mention when workflows change and typically the suggested
+upgrade mechanism is to copy the modified workflows from tuf-on-ci-template.
+
+Supported ways to configure and modify tuf-on-ci workflows:
+* online signing is configured using signing method specific _Repository Variables_,
+  see [ONLINE-SIGNING-SETUP.md] for details
```
lukpueh, line 101: Is the link markup correct?

repo/tuf_on_ci/_repository.py
```diff
@@ -281,9 +281,34 @@ def _validate_role(
         if md.signed.expires > datetime.utcnow() + timedelta(days=days):
             return False, f"Expiry date is further than expected {days} days ahead"
 
+        if isinstance(md.signed, Root):
+            # tuf-on-ci is always consistent_snapshot
+            if not md.signed.consistent_snapshot:
+                return False, "Consistent snapshot is not enabled"
+
+            # Specification: root version must be x+1, not just larger
+            if prev_md and prev_md.signed != md.signed:
```
lukpueh, line 290: Haven't looked at how you're getting `prev_md`, but isn't there a consistency issue if `prev_md.signed == md.signed`?

lukpueh (changes requested):
I haven't tried it out. But from a pure code review this looks great.

lukpueh (approved):

repo/tuf_on_ci/_repository.py
jku: I think there are cases where we check the validity of roles even if the metadata has not (yet) changed in this PR. In that case we wouldn't want to enforce the version bump requirement.

jku:
I'll merge this now, thanks for reviews.

Next steps:
* I'll upgrade jku/tuf-demo to have another test 
* the sigstore root-signing-staging online key is still WIP: I'll setup root-signing-staging once it's available and actually use the custom token there
* once both seem to be working, let's do a tuf-on-ci release and then update the template workflows

PR #159, opened 2024-01-12
repo/signer: Add Signed-Off-By to all commits

jku:
All commits done by signer and repository will now include Signed-Off-By. This makes DCO check happy based on tests in https://github.com/jku/yet-another-test-repo/

I didn't want to make this configurable (because it covers both repository and signer and would be a bit of a pain to manage): hopefully enabling Signed-Off-By for everyone will:
 * not be too annoying for people who don't care about signoff
 * fix the problem for orgs that require it


kommendorkapten (approved):
Looks good. Grep'd the source tree and no git commit was missed.

PR #167, opened 2024-01-19
Add a smoke test for published repository

### LINKED ISSUE #38: feature: watchdog action
There could be an independent failure watchdog action in addition to #37:

* on a cron workflow , run an action that uses a TUF client to fetch the metadata from the published location
* file an issue if refresh fails
* file an issue if online roles are within signing-period: in other words there should have been a version bump already (note that this requires some nuance as online signing happens in cron by default every 6 hours -- so currently we may be 6 hours into the signing period before the signing happens)
* (maybe) file an issue if any delegated offline role has expired
* (maybe) file an issue if offline role is in signing period and a signing event does not exist

jku:
This adds an action that runs a basic client against the repository  

The action could be used in
* cron workflow
* publish workflow

Fixes #38 (at least parts of it)

### DIFF
--- a/actions/test-repository/action.yml
+++ b/actions/test-repository/action.yml
@@ -4,8 +4,10 @@ description: 'Test a published TUF-on-CI repository with a client'
 inputs:
   metadata_url:
     description: 'base metadata URL the client should use'
+    default: ''
   artifact_url:
     description: 'Base artifact URL the client should use'
+    default: ''
 
 runs:
   using: "composite"
@@ -22,6 +24,19 @@ runs:
     - env:
         METADATA_URL: ${{ inputs.metadata_url }}
         ARTIFACT_URL: ${{ inputs.artifact_url }}
+        OWNER_REPO: ${{ github.repository }}
       run: |
+        OWNER=${OWNER_REPO%/*}
+        REPO=${OWNER_REPO#*/}
+
+        # guess reasonable default urls
+        if [ -z $METADATA_URL ]; then
+          METADATA_URL="https://${OWNER}.github.io/${REPO}/metadata/"
+        fi
+        if [ -z $ARTIFACT_URL ]; then
+          ARTIFACT_URL="https://${OWNER}.github.io/${REPO}/targets/"
+        fi
+
+        echo "Testing repository at metadata-url $METADATA_URL, artifact-url $ARTIFACT_URL"
         tuf-on-ci-test-client --metadata-url "$METADATA_URL" --artifact-url "$ARTIFACT_URL"
       shell: bash
--- a/actions/test-repository/action.yml
+++ b/actions/test-repository/action.yml
@@ -8,6 +8,9 @@ inputs:
   artifact_url:
     description: 'Base artifact URL the client should use'
     default: ''
+  expected_artifact:
+    description: 'Optional artifact path that should be checked to exist in the repository'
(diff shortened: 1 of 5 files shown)


jku:
Test run
https://github.com/jku/tuf-on-ci-sigstore-test/actions/runs/7585094401

jku:
> I have no idea if I should expect that to fail very often when we publish to GH Pages and immediately run this...

Based on running a test over the weekend:
* 200 successful publishes and immediate fetches from Pages
* 2 failures, both GitHub outage related
* failures could not have been avoided by retries

 I think I'd be fine adding this in without any retry support: it seems to typically work and #168 is the generics olution to ephemeral failures  -- if issues come up we can add retries or something afterwards.

Mapping out how this can be used in the workflows:
1. We'll want a new workflow anyway: a cron triggered smoke test workflow that calls this action
1. It would be great if the same workflow could be automatically triggered by publishing to pages. I don't think that is possible without another workflow dispatch...
1. So likely publish workflow should just call the same action -- it makes publish more complicated but still readable

jku:
Marked this ready for review: It's a self contained new feature that does not affect users until some workflows start to use the action.

actions/test-repository/action.yml
```diff
@@ -0,0 +1,42 @@
+name: 'Test TUF-on-CI repository'
+description: 'Test a published TUF-on-CI repository with a client'
+
+inputs:
+  metadata_url:
+    description: 'base metadata URL the client should use'
+    default: ''
+  artifact_url:
+    description: 'Base artifact URL the client should use'
+    default: ''
```
kommendorkapten, line 10: How about an optional third parameter that would be a list of targets to download as well? I would think about making it simple, i.e only test for existence, not for verifying content (i.e the message digest).

kommendorkapten:
Another option for robustness is to extract the latest version of the top level metadata and provide that to verify that the correct version is deployed (can be done in a follow up PR).

kommendorkapten (approved):

actions/test-repository/action.yml
jku: Yeah I agree it makes sense: left it out of this version to keep it short but I could do your simplified suggestion...

jku:
> Another option for robustness is to extract the latest version of the top level metadata and provide that to verify that the correct version is deployed (can be done in a follow up PR).

I think the freshness check that is currently in there is nice because you don't have to provide arguments: the argument is essentially the repository ref that we are on and tuf-on-ci-test-client can figure out the correct versions from files on disk.
Currently the freshness test in the end only checks that root and timestamp match the current on-disk versions: this could be expanded if there is a use case this does not cover.

Or did I misunderstand what you meant

kommendorkapten:
No, you are right, I forgot that we updated publish action with an option to take on the git ref. So disregard that comment.

actions/test-repository/action.yml
jku: I've added an optional `expected_artifact` argument. I don't know if it's actually useful for most repos but it'll likely be useful for root-signing and it's not much more complicated.

jku:
I think I'll wait until 0.4 is released before merging: We don't have the workflows ready yet for this so no need to have the action in tuf-on-ci yet either

jku:
merged after 0.4  release

PR #177, opened 2024-01-29
WIP: Add actionlint job

jku:
WIP: I want to see how actionlint works in PRs -- is it useful?


jku:
It looks like actionlint is not actually capable of linting actions, it only works on workflows. 

I feel I am a victim of false advertizing

PR #179, opened 2024-01-31
Switch from black to ruff

jku:
ruff now does formatting as well and it does a decent job:
* We can get rid of one more dependency
* linting now takes a whole 0.28 seconds


kommendorkapten (approved):
🏃

PR #186, opened 2024-02-09
Use PRs instead of issues for signing events

### LINKED ISSUE #78: Use PRs in signing events instead of issues
I originally chose to use issues for signing events because creating incomplete PRs felt wrong -- a pull request implies its ready to be merged. However:
* as @kommendorkapten pointed out it's harder to explain to people who are familiar with GitHub projects
* In any case we need tests that prevent merging if the PR is incomplete so there's no safety issue really
* connection between PR and branch is stronger (and easier to review)
* I don't think there's technically anything that prevents just using PRs instead of issues as signing events

The only thing that's somewhat confusing is that you end up with signature PRs to a branch that is itself a (signing event) PR branch already...

jku:
Pull requests make more sense for the signing event:
* Since there is a branch associated with the signing event, PR matches that better than issue
* The UI is better as the commits are now visible in between the comments

The issue of course is that a signing event can spend a while in a state when it's not ready for merging or even review. This is manageable because:
* we can manage the draft state: the PR is a draft until the signing event is ready
* The check is now visible and useful: it's very clear if the signing event is "green" or not

The main annoyance in this PR is the use of GraphQL to modify draft state. I decided to do that instead of the other options listed in the linked issue discussion. The best alternative would be to rely more on bash and to use `gh` tool as an API. I think I still prefer js to bash for this...

Fixes #78
Example PR: jku/tuf-on-ci-sigstore-test#31

### DIFF
--- a/actions/signing-event/action.yml
+++ b/actions/signing-event/action.yml
@@ -93,16 +93,16 @@ runs:
           }
 
           message = fs.readFileSync('./status-output').toString()
+          summary = "Signing event in progress"
+          should_be_draft = true
           if (process.env.STATUS == 'success') {
             message += "### Signing event is successful\n\n"
             message += "Threshold of signatures has been reached: this signing event can be reviewed and merged."
             summary = "Signing event is successful"
             should_be_draft = false
-          } else {
-            summary = "Signing event in progress"
-            should_be_draft = true
           }
 
+          // Pull request numbers are also valid issue numbers
           github.rest.issues.createComment({
             owner: context.repo.owner,
             repo: context.repo.repo,
--- a/actions/signing-event/action.yml
+++ b/actions/signing-event/action.yml
@@ -99,13 +99,11 @@ runs:
             message += "### Signing event is successful\n\n"
             message += "Threshold of signatures has been reached: this signing event can be reviewed and merged."
             summary = "Signing event is successful"
-            draft = false
+            should_be_draft = false
           } else {
             summary = "Signing event in progress"
-            draft = true
+            should_be_draft = true
           }
-
-          // TODO set draft state here: The rest api does not support that
 
           github.rest.issues.createComment({
             owner: context.repo.owner,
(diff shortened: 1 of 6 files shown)


actions/signing-event/action.yml
```diff
@@ -59,60 +59,83 @@ runs:
...
-          issue = 0
-          const repo = context.repo.owner + "/" + context.repo.repo
-          const issues = await github.rest.issues.listForRepo({
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            labels: [process.env.GITHUB_REF_NAME],
+          const title = `Signing event: ${process.env.GITHUB_REF_NAME}`
+          const repo = `${context.repo.owner}/${context.repo.repo}`
+          const prs = await github.rest.search.issuesAndPullRequests({
+            q: `in:title+"${title}"+state:open+type:pr+repo:${repo}`
```
jku, line 73: * uses `rest.search` because `rest.issues` and `rest.pulls` do not seem to allow filtering pull requests with anything useful.
* uses title instead of label because `rest.pulls.create()` does not support labels. Using the title should be fine: it includes the branch name which we expect to be unique anyway

jku:
Adding reviewers, happy to do a video meeting walkthrough if that helps

actions/signing-event/action.yml
```diff
@@ -59,60 +59,83 @@ runs:
...
-            pr_url.searchParams.set("title", "Signing event " + process.env.GITHUB_REF_NAME)
-            pr_url.searchParams.set("body", "Signing event " + process.env.GITHUB_REF_NAME + " is successful and ready to merge.\n\nCloses #" + issue + ".")
             message += "### Signing event is successful\n\n"
-            message += "Threshold of signatures has been reached. A [pull request](" + pr_url + ") can be opened."
+            message += "Threshold of signatures has been reached: this signing event can be reviewed and merged."
             summary = "Signing event is successful"
+            should_be_draft = false
           } else {
             summary = "Signing event in progress"
+            should_be_draft = true
```
kommendorkapten, line 103: Just my personal preference, ignore if you don't agree. But I would prefer if `should_be_draft = true` is set on line 96 (after `message = fs...`) and then only changed _if_ it should be set to `false`, i.e if event is successful. I just found it a bit strange looking when there is no variable definition in the scope it's later used.

actions/signing-event/action.yml
```diff
@@ -59,60 +59,83 @@ runs:
...
           }
 
           github.rest.issues.createComment({
```
kommendorkapten, line 106: Maybe add a comment that the issues API also works on pull requests, as this line may fool a user? (Every pull request is an issue)

kommendorkapten (approved):
Looks good!

PR #193, opened 2024-02-19
repo: Preserve custom fields in TargetFiles

### LINKED ISSUE #192: artifacts listed as MODIFIED in root-signing
https://github.com/sigstore/root-signing-staging/pull/51

the root-signing artifacts are listed as MODIFIED as 
* tuf-on-ci uses only "sha256" but legacy root-signing has included both sha256 and sha512
* tuf-on-ci does not preserve the custom metadata that legacy root-signing added

Not sure yet what the right plan is here

jku:
This is pretty specific workaround for Sigstore root-signing: The targetfile structure contains custom metadata that would be nice to preserve when the artifact list is re-generated.

Partially fixes #192 

---

There is no test for this (as tuf-on-ci just does not have a way to add this custom metadata) but I've tested it manually

### DIFF
--- a/repo/tuf_on_ci/_repository.py
+++ b/repo/tuf_on_ci/_repository.py
@@ -535,15 +535,20 @@ class CIRepository(Repository):
         if rolename in ["root", "timestamp", "snapshot"]:
             return False
 
-        new_target_dict = self._build_targets(
+        new_tfiles = self._build_targets(
             os.path.join(self._dir, "..", "targets"), rolename
         )
         with self.edit_targets(rolename) as targets:
+            # Keep any existing custom fields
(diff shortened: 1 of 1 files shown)


kommendorkapten (commented):
Would there be updates to the tests to verify this too?

jku:
> Would there be updates to the tests to verify this too?

I'm not sure how to do that without pain... tuf-on-ci doesn't support creating these custom fields so I'd have to manually edit the metadata before signing it.

This code path does get executed in the tests already though: `unrecognized_fields` is just an empty dict in the normal case.

kommendorkapten (approved):

PR #197, opened 2024-02-26
Release 0.7.0

jku:
* Signer has improved signing error handling
* Custom fields in TargetFile metadata are now preserved during target update
  (this is a workaround mostly for sigstore root-signing legacy artifacts)


kommendorkapten (approved):
🚢

PR #203, opened 2024-02-27
Add runtime version check

jku:
Python is a mess with regards to keeping applications updated: Add a runtime message about out-of-date application.

It's a bit complicated but I wanted to not ping pypi constantly if we run the tool several times


jku:
also the version number comparison is a bit suspicious: should probably just depend on `packaging.version` for that

joshuagl (approved):
This is neat!

PR #220, opened 2024-03-15
Support for arbitrarily deep targets and delegates

kipz:
As per the discussion [on cnf slack](https://cloud-native.slack.com/archives/C04SHK2DPK9/p1710266843855449), tuf-on-ci cannot currently support deeply nested targets or delegated role paths.

This PR supports arbitrarily deep target paths (using a '**' glob pattern) ~,but prioritizes the glob patterns in delegated roles (in order in targets.json)~. A delegate role can only "steal" ownership of a top level target if it is first removed from the top level target. Delegated roles are prioritised by order in targets.json if there are overlapping `path` patterns.

Also, the default use of the delegated role name as the the first level directory in the targets directory is removed.

For example:
* If a file isn't covered by any delegated role, it is added to top level targets (by use of '**') glob pattern
* If a delegated role is introduced with a pattern that would include that file ~, the delegate role is chosen to use this file, and it is removed from top level targets~ nothing changes - the file stays owned by the top level targets.
* If a file is removed from the top-level targets (in the normal way), it can then be added to a delegated role.

~Obviously overriding targets from delegated roles could have implications, but this could be handled in a few ways:~

~* Change the implementation so that delegated roles cannot take owner ship of files that were (in the current good metadata) owned by top level targets~
~* Don't use targets and rely entirely on delegated roles~
~* Don't accept PRs that attempt to change targets metadata (they would likely fail to be signed anyway)~


mrjoelkamp:
> If a later (in targets.json) delegated role has a pattern that also matches, it is ignored
It would be good if this threw an error when attempting to create a new delegated role that conflicts with an existing delegated role to prevent it from happening. If it is added manually, then how do you determine which delegated role claimed the target path first? 

> Change the implementation so that delegated roles cannot take owner ship of files that were (in the current good metadata) owned by top level targets
This seems reasonable to me, part of the path conflict resolution above. Don't allow a delegate role to have a path conflict with an existing role. Target files claimed by the top-level targets role would have to be removed and then added back in under a delegated target path.

repo/test/test_repo3/good/metadata/myrole1.json
mrjoelkamp, line 1: Is this the metadata before the target files are discovered? Do they end up getting populated when a test is run? 

I expect to see the delegated target files listed here for this role but not sure at what point in the lifecycle this metadata captured.

kipz:
> It would be good if this threw an error when attempting to create a new delegated role that conflicts with an existing delegated role to prevent it from happening. If it is added manually, then how do you determine which delegated role claimed the target path first?

The reason I didn't do this is because it's allowed in the TUF spec. The client workflow defines the behavior for this scenario AFAICT. But this is certainly an option.

kipz:
> This seems reasonable to me, part of the path conflict resolution above. Don't allow a delegate role to have a path conflict with an existing role. Target files claimed by the top-level targets role would have to be removed and then added back in under a delegated target path.

Yeah, I think I'm leaning in this direction too. Would be good to know if others have thoughts.

repo/test/test_repo3/good/metadata/myrole1.json
kipz: > 
Is this the metadata before the target files are discovered? Do they end up getting populated when a test is run?

Yeah, but the original/good location is left unmodified, only the temporary ones are updated.

The tests do cover this I think, but they don't go as far as the signing flow of course.

kipz:
@mrjoelkamp I've updated the PR to behave as discussed above.

jku:
I've not read the patch yet but I can write down some of the design ideas I have:
* Supporting all of the features in the TUF spec is explicitly **not** a goal: simplicity and maintainability are far more important
* I want to support actual use cases though: I realize tuf-on-ci is currently focused on the sigstore use case
* If TUF spec supported "**" in delegation paths I would absolutely use that
* I'm cautious about _completely arbitrary_ delegation paths: I would like to see good arguments why  we can't enforce that the role DELEGATEDROLE can only delegate paths that begin "DELEGATEDROLE>/...". I think arbitrary delegations could potentially be very confusing
* Even with delegations that are not fully arbitrary but just partly configurable, I'd really like to hear what the use case is that requires more than 3 (or 5, or 10) layers of subdirectories
* I'm asking that because I think there are complexity multipliers here:
   * delegation UX: if delegations are arbitrary then they need to be configurable in `tuf-on-ci-delegate`... What does that look like? do maintainer now have know how to write TUF delegation path patterns? 
   * Do signers now have know how to review arbitrary delegation paths?  The upside of strictly limited delegation paths is that we never have to show the actual delegation patterns to signers -- they see the artifact inside the directory `targets/DELEGATEDROLE/...` and they immediately know who "owns" this artifact
   * artifact UX: this PR presumably manages finding the correct delegated role for the modified artifact during the signing event workflow, but how does  the maintainer know which role the artifact will be added to when they drop the file somewhere in targets/... ? When I imagine a tuf-on-ci maintainer, I think of _a person who does not work on this stuff as their main job but does a bit of tuf-on-ci maintenance maybe every six months_. We should assume they remember almost nothing when they need to do some artifacts changes: we definitely cannot assume they remember arbitrary delegation paths

kipz:
> I've not read the patch yet but I can write down some of the design ideas I have:

Thanks @jku , the code might make your eyes bleed, so let's iterate on the requirements first :)

> * Supporting all of the features in the TUF spec is explicitly **not** a goal: simplicity and maintainability are far more important

Understood, and makes sense.

> * I want to support actual use cases though: I realize tuf-on-ci is currently (too) focused on the sigstore use case

For me, I have a rego bundle layout (a few directories deep) that I'd like to distribute with TUF, and right now, I can't do that using tuf-on-ci.

> * If TUF spec supported "**" in delegation paths I would absolutely use that
> * I'm cautious about _completely arbitrary_ delegation paths: I would like to see good arguments why  we can't enforce that the role DELEGATEDROLE can only delegate paths that begin "DELEGATEDROLE/...". I think arbitrary delegations could potentially be very confusing

Most artifact repository types add constraints on directory naming (for example Maven repos), so needing the role name to match the first level would break this use case. I think this is pretty common.

Having said that, for my use case, I'm happy to have the top level directory match the role name.

> * Even with delegations that are not fully arbitrary but just partly configurable, I'd really like to hear what the use case is that requires more than 3 (or 5, or 10) layers of subdirectories

I can't think of many that would require more than around 5 (maybe maven where group-id is split up by directories?).

> * I'm asking that because I think there are complexity multipliers here:

Maybe we can avoid those, and allow folks to opt-in to the configuration manually if they know their way around?

>   * delegation UX: if delegations are arbitrary then they need to be configurable in `tuf-on-ci-delegate`... What does that look like? do maintainers now have to know how to write TUF delegation path patterns?

Perhaps we can keep the current defaults and not expose that configuration by the CLI/UX. Only if someone manually edits the delegation paths will they ever be exposed to that complexity. Editing patterns isn't that common, but they're pretty straight forward.

>   * Do signers now have to know how to review arbitrary delegation path patterns?  The upside of strictly limited delegation paths is that we never have to show the actual delegation patterns to signers -- they see the artifact inside the directory `targets/DELEGATEDROLE/...` and they immediately know who "owns" this artifact

I think we could probably do a good job at rendering some pretty markdown of the delegation paths and the changes to matching files. And we can of course, we can clearly label which roles they are associated with. Again, this is only if the delegated paths were explicitly opted-in to, the default remaining as is (a single path with the role-name).

>   * artifact UX: this PR presumably manages finding the correct delegated role for the modified artifact during the signing event workflow, but how does  the maintainer know which role the artifact will be added to when they drop the file somewhere in targets/... ? When I imagine a tuf-on-ci maintainer, I think of _a person who does not work on this stuff as their main job but does a bit of tuf-on-ci repo maintenance maybe once every six months_. We should assume they remember almost nothing when they need to do some artifacts changes: we definitely cannot assume they remember arbitrary delegation paths

So I was thinking that by default, the experience of creating new delegations remains unchanged. Let's say it defaults to the role-name as the only path.

But if the user is so inclined, and does understand how to manually edit the delegated role definition to add other paths, then I think it's probably OK for matching artifacts to be picked up and added to the metadata for signing.

This allows the simple/occasional user to work as they currently do, but can opt in to use multiple paths in a delegated role, and deeper trees. Support for manipulating these need not be exposed in the CLI maybe?

Having said all that, my use case only really requires the addition of a small (say 3-5) extra levels to each delegated role, and using the role-name as the top level would work just fine.

We a few options then:
* Constrain delegated paths to N levels
* Only allow roll-name as prefix for delegated paths
* Close for now and collect more use cases

It was a little surprising that when we manually edited the delegated paths that the files weren't picked up, and took some effort to figure out that this just wasn't supported (hence this issue). So if we do want to keep some constraints (tree depth, role-name etc), we could probably add something to the PRs explaining "the following files were not added to the metadata because of the following rules" etc.

What do you reckon?

repo/test/test_repo3/good/metadata/myrole1.json
jku: Later we can also add tests to the the top-level end-to-end tests that handle signing events and signing: they are a bit of a pain to write so should not be done until we're really sure the design is right.

kipz:
@jku @mrjoelkamp thanks for reviewing things.

As discussed @jku , I'll raise another PR that solves our immediate use-case of needing some extra levels, keeping the current path/rolename convention.

PR #237, opened 2024-03-28
signer: Increase signing robustness

jku:
* Make sure we re-initialize the signer when trying to sign again after a failure (this makes sense if e.g. yubikey has been switched midsigning)
* verify the signature after signing
* Only cache signers after a successful signing

### DIFF
--- a/signer/test/test_user.py
+++ b/signer/test/test_user.py
@@ -94,6 +94,9 @@ class TestUser(unittest.TestCase):
                 "762cb22caca65de5e9b7b6baecb84ca989d337280ce6914b6440aea95769ad93",
             )
 
+            # Cache the signer
+            user.set_signer(HSM_KEY, hsm_signer)
+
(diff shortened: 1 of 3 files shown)


jku:
Ok some minimal testing is done:
* wrong signatures are now prevented
* The error messages are not super great, it's pretty much:
  1. "Failed to sign targets with @jku key" (this could be e.g. incorrect PIN but also other failures) 
  1. "Signature fails to verify with @jku key." (we signed but likely the key was incorrect)
  
 I'll spend a moment looking at ways to improve messages but I think this is good to review: it fixes the immediate problem.

jku:
> I'll spend a moment looking at ways to improve messages

* case 1 (_failed to sign_) is improved in next python-tuf release
* I added a suggestion to case 2 (_failed to verify_) error message that this might be the wrong key. We should also store the key details in client cache ... but that's a bit more complicated

kommendorkapten (approved):
🚀

PR #239, opened 2024-03-30
Test repository improvements

jku:
Add multiple optional arguments to the action

This one is for upgrade testing after a signing event:
* Add upgrade testing with `update_base_url`: this allows testing that the new repository is a compliant upgrade from an existing repository version. This is useful when we have a "prod" and "preprod" environments and a  new "preprod" deployment is being tested: in this case "prod" can be used as the `update_base_url`

All of the following are reasonable use cases from sigstore-probers (so out-of-tree testing for alerting purposes):
* Add `compare_source` option to allow running the test without being inside a tuf-on-ci repository (when 'false', the metadata versions are not verified to be specific ones)
* Add `metadata_dir` to allow using the client metadata for e.g. deduplication like sigstore-prober wants to do
* Add `valid_days` to allow verifying if the repository is valid at a specific time in the future
* add `offline_valid_days` to allow verifying that root and targets metadata are valid at  a specific time (even if repository is not)

### DIFF
--- a/repo/tuf_on_ci/client.py
+++ b/repo/tuf_on_ci/client.py
@@ -29,38 +29,49 @@ def expiry_check(dir: str, role: str, timestamp: int):
 @click.option("-v", "--verbose", count=True, default=0)
 @click.option("-m", "--metadata-url", type=str, required=True)
 @click.option("-a", "--artifact-url", type=str, required=True)
+@click.option("-u", "--update-base-url", type=str)
 @click.option("-e", "--expected-artifact", type=str)
 @click.option("--compare-source/--no-compare-source", default=True)
 @click.option("-t", "--time", type=int)
 @click.option("-o", "--offline-time", type=int)
+@click.option("-d", "--metadata-dir", type=str)
 def client(
     verbose: int,
     metadata_url: str,
     artifact_url: str,
+    update_base_url: str | None,
     expected_artifact: str | None,
     compare_source: bool,
     time: int | None,
     offline_time: int | None,
+    metadata_dir: str | None,
 ) -> None:
-    """Test client for tuf-on-ci
-
-    Client expects to be run in tuf-on-ci repository (current metadata
-    in the sources will be considered the expected expected metadata the
-    client should receive from remote)."""
+    """Test client for tuf-on-ci"""
 
     logging.basicConfig(level=logging.WARNING - verbose * 10)
 
     with TemporaryDirectory() as client_dir:
-        metadata_dir = os.path.join(client_dir, "metadata")
+        if metadata_dir is None:
+            metadata_dir = os.path.join(client_dir, "metadata")
         artifact_dir = os.path.join(client_dir, "artifacts")
-        os.mkdir(metadata_dir)
+        os.makedirs(metadata_dir, exist_ok=True)
         os.mkdir(artifact_dir)
 
+        # initialize client with a root.json from metadata_url
         root_url = f"{metadata_url}/1.root.json"
         try:
             request.urlretrieve(root_url, f"{metadata_dir}/root.json")  # noqa: S310
         except OSError as e:
             sys.exit(f"Failed to download initial root {root_url}: {e}")
+
+        if update_base_url is not None:
+            # Update client to update_base_url before doing the actual update
+            updater = Updater(metadata_dir, metadata_url, artifact_dir, artifact_url)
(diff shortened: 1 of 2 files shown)


jku:
This actually looks quite nice (some of the tests are expected to fail, that's why it's red):
https://github.com/jku/test-tuf-on-ci-tests/actions/workflows/test-tuf-on-ci.yml

jku:
I think the only issue left is that client initialization now happens with 1.root.json from the repository-under-test. This is likely a problem for prod sigstore root-signing as the early metadata is not compliant.

I didn't want to add yet another input arg but maybe it's required

repo/tuf_on_ci/client.py
```diff
@@ -5,74 +5,110 @@
...
-                if os.path.isfile(fpath):
-                    shutil.copy(fpath, os.path.join(metadata_dir, fname))
-        else:
-            # pre-populate the client cache with first root version from source
-            root = "metadata/root_history/1.root.json"
-            shutil.copy(root, os.path.join(metadata_dir, "root.json"))
-
-        # For now, just confirm we can get top level metadata from remote
+        # initialize client with a root.json from metadata_url
+        root_url = f"{metadata_url}/1.root.json"
```
kommendorkapten, line 62: If we want to support an option to start at a given version of the root (like 5 for sigstore), how about an option that specifies the starting version that is used here (could be simple implement as the default argument can be `1`), which can be used here? It may not be as complicated as an new option for the starting root url?

repo/tuf_on_ci/client.py
```diff
@@ -5,74 +5,110 @@
...
+        try:
+            request.urlretrieve(root_url, f"{metadata_dir}/root.json")  # noqa: S310
+        except OSError as e:
+            sys.exit(f"Failed to download initial root {root_url}: {e}")
+
+        if update_base_url is not None:
+            # Update client to update_base_url before doing the actual update
+            updater = Updater(metadata_dir, metadata_url, artifact_dir, artifact_url)
+            try:
+                updater.refresh()
```
kommendorkapten, line 72: Is the intended purpose of `update_base_url` to only verify that it can update with the new repo, and then continue below to perform the tests against the configured url.

I'm thinking if this needs clarification when e.g. both `update_base_url` and `validity` is provided, is it clear which repo the time validity is tested against?

kommendorkapten: Wait, I think i understand now. This is only used to load the cache on disk, then below on line 77 we update again to get the fresh metadata, which we actually test again, right?

jku: Yes that sounds 100% correct:
* metadata_url is always what were really testing: so `valid_days` etc are always checked on data from that URL
* normally we only ensure that the refresh works if local cache contains 1.root.json
* if update_base_url is specified we first refresh from that URL (to populate local cache), then ensure refresh works: this makes sure e.g. snapshot contents don't illegally change between the two repos

kommendorkapten: Perfect!

jku: oh wait you found a bug -- it uses the wrong URL :joy:

kommendorkapten: oh yes, it should be `update_base_url`.

jku: fixed now, thanks!

repo/tuf_on_ci/client.py
jku: I implemented this using "root.json" from working directory: this way the workflow can provide the file in whatever way they want

jku:
I htink this looks decent, here's a test with actual sigstore prod (so it needs to initialize with v5 root):
https://github.com/jku/test-tuf-on-ci-tests/actions/runs/8536137266/job/23384039025

repo/tuf_on_ci/client.py
jku: it's not verifiable in the action output but with debug output once can see that the correct url is getting used:

`tuf-on-ci-test-client -vvv -m https://sigstore.github.io/root-signing-staging/ -a https://tuf-repo-cdn.sigstage.dev/targets/  --no-compare-source --update-base-url https://tuf-repo-cdn.sigstage.dev/

`

kommendorkapten (approved):
Nice!

PR #243, opened 2024-04-03
repo, signer: Pin securesystemslib better

jku:
Securesystemslib is "pre 1.0" so the normal pinning does not really work: the new pinning "~= 0.31.0" should lead to 0.31.1 to be accepted but not 0.32.0.


lukpueh (approved):
According to the [docs](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release) this means `>= 0.31.0, == 0.31.*`, which makes sense.

LGTM

PR #248, opened 2024-04-05
Pin all the deps

jku:
### Pin all (even indirect) dependencies when running `repo/` in GH actions
* `requirements.txt` is rebuilt by running pip-compile on `repo/pyproject.toml`
* this build happens in a new workflow that runs once a week and creates a PR if something changed
* requirements.txt is used as a constraint when installing `repo/` within the actions

Some drawbacks of this approach:
* We're doing this manually like some kind of cavepeople: bugs are inevitable in 70 lines yaml
* The reason for doing it manually is that as far as I can tell dependabot cannot manage adding/removing indirect dependencies as needed. As a result the PR lacks all links and details that Dependabot PRs have
* We _could_ run this in addition to running Dependabot on requirements.txt: this way normal upgrades could be handled by dependabot and the caveman approach would be used for added/removed deps -- the downside here is dealing with potentially duplicate PRs
* pip is difficult so hashes are not used in this version (you can't use hashes as constraints if the thing you're installing does not have hashes)...

If I should be using a tool that would make this easier please let me know.

CC @lukpueh @woodruffw in case you want to have a look since I complained to both of you about this...

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,7 +20,7 @@ jobs:
           python-version: '3.11'
 
       - name: Install build dependencies
-        run: python3 -m pip install --constraint build-constraints.txt build
+        run: python3 -m pip install -c build-constraints.txt build
 
       - name: Build binary wheel and source tarball
         run: PIP_CONSTRAINT=build-constraints.txt python3 -m build --sdist --wheel --outdir dist/ signer/
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,10 +20,10 @@ jobs:
           python-version: '3.11'
 
       - name: Install build dependencies
-        run: python3 -m pip install build
+        run: python3 -m pip install --constraint build-constraints.txt build
 
       - name: Build binary wheel and source tarball
-        run: python3 -m build --sdist --wheel --outdir dist/ signer/
+        run: PIP_CONSTRAINT=build-constraints.txt python3 -m build --sdist --wheel --outdir dist/ signer/
 
       - name: Store build artifacts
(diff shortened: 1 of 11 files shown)


.github/workflows/update-pinned-deps.yml
```diff
@@ -0,0 +1,74 @@
...
+
+    steps:
+      - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2
+
+      - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
+        with:
+          python-version: "3.11"
+
+      - name: Install pip-tools
+        run: pip install pip-tools
```
jku, line 28: this should be version constrained in another requirements file... but so should build and hatchling and any other build deps: I'll handle that in another PR

kommendorkapten:
It's a bit of extra work to achieve this, but I don't see a better options. Would be nice with a native solution that tracked all dependencies.

.github/workflows/release.yml
```diff
@@ -20,10 +20,10 @@ jobs:
           python-version: '3.11'
 
       - name: Install build dependencies
-        run: python3 -m pip install build
+        run: python3 -m pip install --constraint build-constraints.txt build
```
kommendorkapten, line 23: nit: for consistency, use either short or long flags? (ci.yaml uses `-c`)

kommendorkapten (approved):
Looks good, only one nit comment

jku:
Well, let's see what happens... the constraints file should get the initial PR when this is merged.

PR #262, opened 2024-04-12
dependabot: Bring back individual configs for signer/repo

jku:
It seems like *.txt files are found anywhere under the configured dependabot dir but pyproject.toml are found only in the configured dir itself.

Add separate configs to get updates to pyproject.tomls.


jku:
Modified this to use the dependabot beta feature `directories` as well

kommendorkapten (approved):

PR #294, opened 2024-04-29
signer: Make sure initial keyid is spec-compliant

jku:
Spec demands that keyids be calculated from the keys canonicalized json. This is not that useful and it is a bit of a pain in the butt so most clients don't bother... but it is in the spec.

Make sure the keyids we generate are compliant.

### DIFF
--- a/signer/tuf_on_ci_sign/_signer_repository.py
+++ b/signer/tuf_on_ci_sign/_signer_repository.py
@@ -14,6 +14,8 @@ from glob import glob
 
 import click
 from securesystemslib.exceptions import UnverifiedSignatureError
+from securesystemslib.formats import encode_canonical
+from securesystemslib.hash import digest
 from securesystemslib.signer import (
     KEY_FOR_TYPE_AND_SCHEME,
     SIGNER_FOR_URI_SCHEME,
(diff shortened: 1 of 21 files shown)


jku:
I suppose this is ready for review:
* existing keys are not modified
* whenever a custom field is added to a key, keyid is updated: in practice this happens right after a new key is imported/generated
* existing repositories need to "rotate" keys if they want to fix the issue: In practice it's likely easiest to remove keys, then add them as "new" ones

signer/tuf_on_ci_sign/_signer_repository.py
```diff
@@ -97,6 +99,15 @@ def _find_changed_roles(known_good_dir: str, signing_event_dir: str) -> list[str
     return changed_roles
 
 
+def set_key_field(key: Key, name: str, value: str) -> None:
+    key.unrecognized_fields[f"x-tuf-on-ci-{name}"] = value
+    # This is dumb but TUF spec requires the keyid to change at this point
```
lukpueh, line 104: This is dumb indeed and ambiguous too. As one could argue that unrecognised fields are not part of the "canonical form" of the key.

lukpueh (approved):
Looks good and is probably complete (I grepped for `unrecognized_fields.*=`)

kommendorkapten (approved):

jku:
Thanks. I'll still take a moment tomorrow to think through fredriks idea (to see if tuf-on-ci-signer would be capable of "fixing" the problem in existing root-signing-staging repo in a single signing event): I think that can be in a different PR but I'll have a look first

jku:
by complete accident I think the signer might actually work correctly out-of-the box: I'll write some code that updates all the keyids in one go to test this...

jku:
I will merge this and file a new one for the "fix-noncompliant-keyids" feature: I've got a branch but it's big enough that a new PR is a good idea

PR #296, opened 2024-04-30
Use dependabot beta features

jku:
This is a public beta feature that should help with dependabot spam for the actions by combining all actions (and workflows) updates into one issue

https://github.blog/changelog/2024-04-29-dependabot-multi-directory-configuration-public-beta-now-available/

--- 

I'd like to try the beta since they're asking for feedback. We should see if it's working immediately as I'm also adding the last two actions so that will trigger some updates...

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,33 +31,13 @@ updates:
   schedule:
     interval: "weekly"
 
-
 - package-ecosystem: "github-actions"
-  directory: "/"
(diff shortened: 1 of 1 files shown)


kommendorkapten (approved):

jku:
The beta looks buggy (as others have reported already) so all updates are not happening... we probably also want to group the updates together. since they don't seem to automatically  get grouped

PR #310, opened 2024-05-07
dependabot: Group the actions updates

jku:
At least currently Dependabots new "directories" feature only creates one issue at a time (even if it should file multiple). I hope this grouping will at least include all of the changes in a single PR.


lukpueh (approved):
Let's see what happens.

PR #313, opened 2024-05-13
repository: Add Markdown description

jku:
This is a very initial implementation of human readable repository state
* tuf-on-ci-build-repository now creates a markdown file in the metadata publishing directory
* markdown contains some information on the repository contents
* there is a jekyll-build step that creates html and css from that markdown 

There are some downsides to this PR:
* The content is likely not 100% complete (e.g. artifacts are not listed at all): wanted to to keep the PR to a decent size
* a dynamic js  solution (where a separate js+html just reads the current metadata and renders it for humans) would be better but I don't have the skills: this can be replaced later with something more professional later. The advantages would be 
   * could have dynamic content like _"signing event starts in 2 days"_
   * tuf-on-ci-build-repository would not need to generate anything and the jekyll dependency would not be needed
* adding markdown, html and css into published metadata directory is a bit annoying but putting them elsewhere requires rethinking the tuf-on-ci-build-repository CLI "api". I suppose this location should be configurable as well

See https://jku.github.io/tuf-on-ci-sigstore-test/metadata/ for an example.


repo/tuf_on_ci/build_repository.py
```diff
@@ -3,14 +3,97 @@
...
+        *cmd,
+    ]
+    proc = subprocess.run(cmd, check=True, capture_output=True, text=True)
+    logger.debug("%s:\n%s", cmd, proc.stdout)
+    return proc
+
+
+def build_description(repo: CIRepository) -> str:
+    lines = [
+        "## Signers",
```
jku, line 36: This title is not great. Maybe _Repository state_ or nothing at all (considering that jekyll will add the project name on top anyway)

jku:
TODO: Add the markdown to e2e expected results -- will have to consider if we really want to byte by byte compare the markdown...

repo/tuf_on_ci/build_repository.py
```diff
@@ -3,14 +3,95 @@
...
+        if rolename == "timestamp":
+            json_link = f"{rolename}.json"
+        else:
+            json_link = f"{delegate.version}.{rolename}.json"
+        expiry = delegate.expires
+        signing_days, _ = repo.signing_expiry_period(rolename)
+        signing = expiry - timedelta(days=signing_days)
+        name_str = f"{rolename} ([json]({json_link}))"
+        date_str = f"[Starts {signing.strftime('%Y-%m-%d')}](## '{signing} - {expiry}')"
+        signer_str = f"{role.threshold} of {', '.join(signers)}"
```
joshuagl, line 73: minor nit: my brain stumbled over this "foo, bar, baz of X" phrasing a few times. What about "foo, bar, baz: X required" ?
```suggestion
        signer_str = f"{', '.join(signers)}: {role.threshold} required"
```

joshuagl (approved):
This is great. Simple, but effective. I've made one minor suggestion, but really this is fine as is.

repo/tuf_on_ci/build_repository.py
```diff
@@ -3,14 +3,95 @@
...
+
+    for delegator, rolename in roles:
+        role = delegator.get_delegated_role(rolename)
+        keys = [delegator.get_key(keyid) for keyid in role.keyids]
+        signers = []
+        for key in keys:
+            if "x-tuf-on-ci-keyowner" in key.unrecognized_fields:
+                signers.append(key.unrecognized_fields["x-tuf-on-ci-keyowner"])
+            else:
+                signers.append("_online key_")
```
lukpueh, line 61: matter of taste, but this could be: 

```suggestion
            keyowner = key.unrecognized_fields.get("x-tuf-on-ci-keyowner", "_online key_")
            signers.append(keyowner)
```

lukpueh (approved):
very cool

repo/tuf_on_ci/build_repository.py
jku: Agreed. I made that _"firstsigner, secondsigner, thirdsigner (2 of 3 required)"_

PR #330, opened 2024-05-15
Refactor dependency management once more

jku:
Dependabot still bundles dependencies I don't want bundled:
* Move build constraints to /build/ so this dependabot group will not find action constraints (previously in /repo/install)
* Handle tox with the build constraints (it's a test dependency but used in workflows like the other build constraints
* Move the action constraints from /repo/install/requirements.txt to /action-constraints.txt: This way dependabot should not see it
* Make actions updates weekly as well: Daily updates are too much


jku:
Lukas if you can take a look that's appreciated: I'm trying to avoid this: https://github.com/theupdateframework/tuf-on-ci/pull/328 -- that PR should only update the minimum versions in pyproject.toml but it gobbles up the pinned constraints too (which I'd rather handle with pip-compile in update-pinned-dependencies.yml) 

The issue seems to be that dependabot finds "*.txt" files _anywhere_ under the specified directory and updates them -- so the constraints need to live outside of any dependabot managed directories

.github/dependabot.yml
```diff
@@ -33,8 +18,13 @@ updates:
   schedule:
     interval: "weekly"
   groups:
-    pyproject-dependencies:
-      # Dependency ranges set in {signer,repo}/pyproject.toml
+    pinned-test-dependencies:
+      # Dependencies pinned {signer,repo}/pyproject.toml to ensure test reproducibility
```
lukpueh, line 22:
```suggestion
      # Dependencies pinned in {signer,repo}/pyproject.toml to ensure test reproducibility
```
?

.github/dependabot.yml
```diff
@@ -48,8 +38,7 @@ updates:
     - "/actions/update-issue/"
     - "/actions/upload-repository/"
```
lukpueh, line 39: Unrelated, but may I ask why all `actions/` subdirs are listed explicitly? Given the findings (Dependabot recursing into subdirs) that led to the current PR, would `/` or `/actions` be enough?

lukpueh (approved):
Looks good. This would be easier, if dependabot had an `exclude-manifest` option, preferably on the group level.

The only thing I don't like is the name `action-constraints.txt`. Because they are also used by tox locally. What about something like `test-constraints.txt`, or `pinned-runtime-constraints.txt`, ... 

Or, if that litters the directory root, you could also move both  constraints files into a common subdirectory, e.g:
req/build/constraints.txt
req/test/constraints.txt

Or you just leave it as it is :D

jku:
> The only thing I don't like is the name action-constraints.txt. Because they are also used by tox locally. What about something like test-constraints.txt, or pinned-runtime-constraints.txt, ...

It could be `pinned-repo-runtime constraints.txt`... but even then the real reason for pinning is that I believe the *actions* should not upgrade dependencies silently, so action-constraints kind of makes sense to me at least. The same constraints are used in tests only because the constraints happen to be available -- the reason for the constraints is still that I want to keep the actions stable.

Issue #292, opened 2024-04-26
keyids for new keys are not spec compliant


jku:
* Spec currently demands that keyids match sha256 of the canonicalized key content and that clients verify that
 * This is widely regarded as a mistake (see e.g. [TAP 12](https://github.com/theupdateframework/taps/blob/master/tap12.md)) and the verification is typically not implemented in clients but the requirement is still in the spec... and apparently at least tuftool from tough project does the verification
 * tuf-on-ci-sign currently 
   * generates a new key: this assigns a keyid that is correct
   * later inserts a custom field (x-tuf-on-ci-keyowner or x-tuf-on-ci-online-uri) into the key : this makes the keyid incorrect WRT spec

jku:
Fixing the initial key creation in tuf-on-ci seems doable: make a wrapper for modifying the custom fields, have the wrapper also re-calculate the keyid

There are a couple of issues though:
* "fixing" existing tuf-on-ci metadata seems tricky: we essentially want to do a key rotation from keyid 1 to keyid 2 where both keyids refer to same underlying key... but this is also explicitly banned by the spec
* in the future, any changes to the fields of the key json (like an online key url changing) would require keyid change

jku:
The alternative "solution" is to push for a spec change: 
* the requirement to verify keyid calculation seems to have no security benefits
* most clients do not implement it
* it adds one more complication to client workflow (which should be simpler, not more complicated) 
* it clearly seems to be a head ache for repository maintenance

kommendorkapten:
This is quite unfortunate, and as @jku points out, having a good set of conformance tests would have helped us here, as apparently almost no client is verifying the key ids so this bug didn't get surfaced.

jku:
Recap:
* Some clients enforce the keyid calculation which means they will not accept repositories created before this fix
* After these fixes, new keys are now created with compliant keyids
* import-repository now changes keyids so they stay compliant
* a `--force-compliant-keyids` flag was added to `tuf-on-ci-delegate`: This can be used to create  a signing event
 that fixes already published root and targets metadata

PR #338, opened 2024-05-17
Fix noncompliant keyids

### LINKED ISSUE #336: signer: add feature to fix non-compliant keyids
this is related #292

jku:
Fixes #336 , relates to #292 

This change will ensure that a a keyid fix can be done in a single signing event

* Make sure signing works with multiple keys (the UI for identifying which key is used isn't there yet but that's ok since so far this will only be used for the non-compliant keyid fix where the actual signing key is the same one, just with two different keyids)
* Add a hidden `--force-compliant-keyids` flag to delegate: this will update all keyids in the role and will mark all affected delegations to be resigned

This approach (just switching the all keyids) looks surprising but  should be 100% spec compliant. Example:
* root v1 has one root signer with key A using non-compliant keyid _abcd_. The metadata signatures contains a sig from _abcd_
* `tuf-on-ci-delegate --force-compliant-keyids sign/fix-keyids root` is used: the signing event now contains a root v2 with one root signer with key A using valid keyid _efgh_ (same key, different keyid)
* the signer runs `tuf-on-ci-sign sign/fix-keyids` as usual: this signs twice with A; The metadata signatures now contain a sig from _abcd_ and a sig from _efgh_
* when client see root v2 they will ensure it is signed by old signers (_abcd_) and new signers (_efgh_). The fact that this happens to be the same key should not be an issue (since the keys are part of two different threshold calculations)

### DIFF
--- a/signer/tuf_on_ci_sign/_signer_repository.py
+++ b/signer/tuf_on_ci_sign/_signer_repository.py
@@ -742,24 +742,26 @@ class SignerRepository(Repository):
     def sign(self, rolename: str):
         """Sign without payload changes"""
         md = self.open(rolename)
+        signing_keys: dict[str, Key] = {}
         for key in self._get_keys(rolename):
             keyowner = key.unrecognized_fields["x-tuf-on-ci-keyowner"]
             if keyowner == self.user.name:
-                self._sign(rolename, md, key)
-                self._write(rolename, md)
-                return
+                signing_keys[key.keyid] = key
 
-        # Root is eligible to sign current root if the signer was valid
+        # user is also eligible to sign current root if the signer was valid
         # in previous version
         if rolename == "root":
             for key in self._get_keys(rolename, True):
                 keyowner = key.unrecognized_fields["x-tuf-on-ci-keyowner"]
                 if keyowner == self.user.name:
-                    self._sign(rolename, md, key)
-                    self._write(rolename, md)
-                    return
+                    signing_keys[key.keyid] = key
 
-        raise ValueError(f"{rolename} signing key for {self.user.name} not found")
+        if not signing_keys:
+            raise ValueError(f"{rolename} signing key for {self.user.name} not found")
+
+        for key in signing_keys.values():
+            self._sign(rolename, md, key)
+        self._write(rolename, md)
 
 
 def build_paths(rolename: str, depth: int) -> list[str]:
--- a/signer/tuf_on_ci_sign/_signer_repository.py
+++ b/signer/tuf_on_ci_sign/_signer_repository.py
@@ -2,6 +2,7 @@
 
 """Internal repository module for TUF-on-CI signer tools"""
 
+import copy
 import filecmp
 import json
 import logging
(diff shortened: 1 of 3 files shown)


jku:
I'm still going to do a bit more manual testing but I think this will work

kommendorkapten:
Code looks reasonable! Waiting for more test results.

jku:
awslabs/tough client does now "succeed" in that root is considered valid but 

```
$ tuftool download --root 3.root.json -m https://jku.github.io/tuf-demo/metadata/ -t https://jku.github.io/tuf-demo/targets -n file1.txt out/`
Failed to load repository: Failed to parse timestamp metadata: missing field `length` at line 14 column 4
```

This one I'm not as inclined to "fix" in tuf-on-ci. I don't know if my decision to avoid lengths and hashes in timestamp and snapshot is the best choice but both options seem to have upsides and downsides and both are spec compliant

jku:
@kommendorkapten I'm done I think: As far as I can tell the code is good. and the results metadata update is solid. awslabs/tough tuftool still doesn't fully work, but the specific issue of noncompliant keyids seems to gets solved by this code

jku:
Hmm, actually tuf-on-ci-repo-import could now also modify the keyids immediately when the custom metadata is added... I'll try to add this in this PR.

jku:
> Hmm, actually tuf-on-ci-repo-import could now also modify the keyids immediately when the custom metadata is added... I'll try to add this in this PR.

the change is simple but I think I have to rerun the root-signing import test to verify this

jku:
> Hmm, actually tuf-on-ci-repo-import could now also modify the keyids immediately when the custom metadata is added... I'll try to add this in this PR.

There is a complication in the root case:
* the import works nicely if the signers for root N and N+1 are the same (tuf-on-ci can only know who the N+1 signers are but this does not matter because the signers are the  same)
* the keyid fix with `tuf-on-ci-delegate --force-compliant-keyids` also work nicely since from tuf-on-ci perspective the N and N+1 signers are completely different: this leads to two signatures for each signer whose keyids was modified
* combining these two is an issue: how can `tuf-on-ci-sign` figure out it needs to sign for two keyids when both import and keyid fix have been done?

jku:
Example import signing event https://github.com/jku/root-signing-test/pull/16
* this is a root-signing lookalike where I am signer for everything
* note how the keys remain same, but keyids change because the keys custom metadata changed
* the old keyid still signs root because of the special case code

kommendorkapten:
> combining these two is an issue: how can tuf-on-ci-sign figure out it needs to sign for two keyids when both import and keyid fix have been done?

Maybe a hack but what if tuf-on-ci during *import* creates an in-memory representation of the yubikey using the "old format", i.e without `x-tuf-on-ci-keyowner`, and compares with the current root, and if a match, when signing it writes the signature to both entries?

kommendorkapten (approved):

jku:
> Maybe a hack but what if tuf-on-ci creates an in-memory representation of the yubikey using the "old format", i.e without `x-tuf-on-ci-keyowner`, and compares with the current root, and if a match, when signing it writes the signature to both entries?

We discussed this live but documenting for others: this is pretty much what the current PR does. For each signer it takes the "tuf-on-ci managed key", calculates keyid without the custom metadata, and looks for keys with that keyid in the non-managed (N-1)  metadata. These keys will then also be used to sign.

PR #349, opened 2024-05-27
Release v0.10

jku:
* Add some automation to build GH changelog
* bump versions and add changelog entry


.github/workflows/release.yml
```diff
@@ -22,14 +22,18 @@ jobs:
       - name: Install build dependencies
         run: python3 -m pip install -c build/build-constraints.txt build
 
-      - name: Build binary wheel and source tarball
-        run: PIP_CONSTRAINT=build/build-constraints.txt python3 -m build --sdist --wheel --outdir dist/ signer/
+      - name: Build release changelog, signer wheel & source tarball
```
kommendorkapten, line 25: Although we don't have a good way (yet) to verify the wheel, how do you feel about adding build provenance for that? After the file is built we could attest it with https://github.com/actions/attest-build-provenance

jku: without a real verification story (who verifies, how), I'm not super invested. I'll review a PR though.

kommendorkapten (approved):

PR #352, opened 2024-05-28
repo: Handle root correctly in create-signing-events

### LINKED ISSUE #347: periodic root bump does not modify root_history
Modifying root using tuf-on-ci-delegate leads to also creating a copy of the file in root_history/.
It seems this does **not** happen when create-signing-event creates a new version of root but should. This is not critical as the first signing will include the missing file... but it creates commits that are a bit unobvious.

I think this just requires `CIRepository.close()` to handle this special case: after writing the actual file, also write  `f"root_history/{version}.{role}.json"`.

jku:
The versioned file in `metadata/root_history/` should get created when root.json is modified.

Also add a end to end test for create-signing-events.

Fixes #347

### DIFF
--- a/repo/tuf_on_ci/_repository.py
+++ b/repo/tuf_on_ci/_repository.py
@@ -191,6 +191,18 @@ class CIRepository(Repository):
 
         return (signing_days, expiry_days)
 
+    def _write(self, rolename: str, md: Metadata) -> None:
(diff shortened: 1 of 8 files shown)


joshuagl (approved):
Changes LGTM.

jku:
thanks for reviews!

PR #358, opened 2024-06-04
signer: Fix keyid tweaking in import-repo

jku:
We only want to fix keyids when the import is actually possible (not when we're just printing out the import json template)


kommendorkapten (approved):
yes, makes sense!

PR #377, opened 2024-07-15
Increase the number of root rotations allowed during TUF refresh.

kommendorkapten:
The default value of 32 is too low for some TUF repositories.


jku (approved):
Looks good

jku (commented):
Wait, do you not need to import?

jku (changes requested):
> tuf_on_ci/client.py:86:18: F821 Undefined name `UpdaterConfig`

yeah import is needed (there's also a line length issue)

jku (approved):

Issue #75, opened 2023-09-06
feature: support external delegated targets


jku:
From sigstore-rootsigning meeting: The registry.npmjs.org role is 
1. managed externally, outside of the repository 
2. actually signed by a KMS key

We should make sure we either support this setup or figure out how to fulfill the  requirements otherwise:
* I think  _targets role signed by online keys_ in general is a good feature
* some internal or external process creating a signing event branch with a fully signed delegated targets change should "just work" already (the idea here is that the merging requires maintainer approval even if the key is online, because the online process is external... but I can see an argument for just merging automatically)
* creating the delegation needs some thinking -- 
  * probably we could just add KMS choice to the `tuf-on-ci-delegate` UX, but then the KMS would need to add the signature to the signing event branch? Or we need to allow creating a delegation without the roles metadata
  * alternatively adding the delegation is also expected to happen externally (and just magically appears in a signing event branch): we just make sure online keys are allowed in the repository code.
 * this would need to be documented well -- that the actual maintenance of the delegated role is not currently handled by tuf-on-ci

jku:
After some thinking:
* normal signing events should indeed "just work" if the external signer is just another signer from the repositorys perspective
* when an update to the delegated target is needed:
  * external automation (or a human) creates the new metadata version , commits it, and pushes the commits to a signing event branch on the repository
  * signing event automation sees the branch, creates an issue and updates the event status: it should now be fully signed and ready to merge
* delegation changes are trickier but they can be considered a UX problem: 
  * we need to allow "importing" KMS keys as (offline) signers
  * but unlike online KMS keys we probably don't want to embed the  the private key uri (`x-tuf-on-ci-online-uri`) in the key but instead want to use `x-tuf-on-ci-keyowner` -- for the repository this is not a online key but just a signer that uses signing events like all the other signers?

kommendorkapten:
> but unlike online KMS keys we probably don't want to embed the the private key uri (x-tuf-on-ci-online-uri) in the key but instead want to use x-tuf-on-ci-keyowner -- for the repository this is not a online key but just a signer that uses signing events like all the other signers?

Yes, that makes a lot of sense.

[...]

PR #384, opened 2024-07-19
Added support for delegations using cloud KMS

kommendorkapten:
This is mainly done by modifying tuf-on-ci-delegate to accept an online key as a signer for delegations. This require some extra care when dealing with the signer's identity. Keyowner or online-uri is now used.

Tuf-on-ci-online-sign now supports an extra parameter --role. This is used to perform online signing with delegations.

Closes https://github.com/theupdateframework/tuf-on-ci/issues/75

### DIFF
--- a/actions/online-sign-targets/action.yml
+++ b/actions/online-sign-targets/action.yml
@@ -103,7 +103,7 @@ runs:
     - name: Dispatch another signing event workflow
       # dispatch if using default token: otherwise update_targets step
       # has alreaddy triggered a push event
-      if: inputs.token == github.token && steps.update_targets.outputs.targets_signed == 'true'
+      if: inputs.token == github.token && steps.sign_targets.outputs.targets_signed == 'true'
       uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
       with:
         github-token: ${{ inputs.token }}
--- a/actions/online-sign-targets/action.yml
+++ b/actions/online-sign-targets/action.yml
@@ -1,25 +1,47 @@
-name: 'Signing event'
-description: 'TUF-on-CI Signing event management'
+name: 'Online signing of targets'
+description: 'TUF-on-CI Online signer for targets'
 
-# This action is called from signing-event workflow, which is dispatched in multiple ways
-# depending on the type of token provided in inputs.
-#
-# 1. create-signing-events action creates a new signing event branch
-#     * When using a custom token this triggers push event handler
-#     * When using the default GitHub token the action calls createWorkflowDispatch()
-# 2. A signer pushes artifact changes to a signing event branch
-#     * This triggers push event handler
-# 3. This action (signing-event) makes a metadata change in update_targets step as a result of an artifact change
-#     * When using a custom token this triggers push event handler
-#     * When using the default GitHub token the action calls createWorkflowDispatch()
-#
-# Cases 1 & 3 lead to status step running. Case 2 leads to skipping status step, but
-# triggering case 3 immediately afterwards.
+# This action is not called from any standard workflows.
+# To use this, create custom workflow that triggers this action when
+# needed. This is typically done for delegations, that would be
+# triggered based on a change to the delegation metadata file.
 
 inputs:
   token:
     description: 'GitHub token'
     required: true
+  targets_to_sign:
+    description: "whitespace separated list of targets role names that should be signed with KMS"
+    required: false
+    default: ""
+  gcp_workload_identity_provider:
+    description: "Google Cloud workload identity provider (required if GCP is used to sign targets roles)"
+    required: false
+    default: ""
+  gcp_service_account:
+    description: "Google Cloud service account name (required if GCP is used to sign targets roles)"
+    required: false
+    default: ""
+  aws_region:
+    description: "AWS region"
+    required: false
+    default: ""
+  aws_role_to_assume:
+    description: "AWS role to assume"
+    required: false
+    default: ""
+  azure_client_id:
+    description: "Azure SPN client id (required to use Azure to sign target roles)"
+    required: false
+    default: ""
+  azure_tenant_id:
+    description: "Azure SPN tenant id (required to use Azure to sign target roles)"
+    required: false
+    default: ""
+  azure_subscription_id:
+    description: "Azure SPN subscription id (required to use Azure to sign target roles)"
+    required: false
+    default: ""
 
 runs:
   using: "composite"
@@ -28,6 +50,29 @@ runs:
       with:
         token: ${{ inputs.token }}
         fetch-depth: 0
+
+    - name: Authenticate to Google Cloud
+      if: inputs.gcp_workload_identity_provider != ''
+      uses: google-github-actions/auth@71fee32a0bb7e97b4d33d548e7d957010649d8fa # v2.1.3
+      with:
+        token_format: access_token
+        workload_identity_provider: ${{ inputs.gcp_workload_identity_provider }}
+        service_account: ${{ inputs.gcp_service_account }}
+
+    - name: Authenticate to AWS
+      if: inputs.aws_role_to_assume != ''
+      uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 #v4.0.2
+      with:
+        aws-region: ${{ inputs.aws_region }}
+        role-to-assume: ${{ inputs.aws_role_to_assume }}
+
+    - name: Authenticate to Azure cloud
+      if: inputs.azure_client_id != ''
+      uses: azure/login@6c251865b4e6290e7b78be643ea2d005bc51f69a # v2.1.1
+      with:
+        client-id: ${{ inputs.azure_client_id }}
+        tenant-id: ${{ inputs.azure_tenant_id }}
+        subscription-id: ${{ inputs.azure_subscription_id }}
 
     - uses: actions/setup-python@39cd14951b08e74b54015e9e001cdefcf80e669f # v5.1.1
       with:
@@ -40,20 +85,23 @@ runs:
         echo "::endgroup::"
       shell: bash
 
-    - id: update_targets
+    - id: sign_targets
+      if: inputs.targets_to_sign != ''
+      env:
+        TARGETS_TO_SIGN: ${{ inputs.targets_to_sign }}
       run: |
-        if tuf-on-ci-update-targets >> update-output;  then
-          echo "targets_updated=true" >> $GITHUB_OUTPUT
+        if tuf-on-ci-online-sign-targets $TARGETS_TO_SIGN >> sign-output;  then
+          echo "targets_signed=true" >> $GITHUB_OUTPUT
         else
-          echo "targets_updated=false" >> $GITHUB_OUTPUT
+          echo "targets_signed=false" >> $GITHUB_OUTPUT
         fi
-        cat update-output
-        cat update-output >> output
-        cat update-output >> "$GITHUB_STEP_SUMMARY"
+        cat sign-output
+        cat sign-output >> output
+        cat sign-output >> "$GITHUB_STEP_SUMMARY"
       shell: bash
 
     - id: status
-      if: steps.update_targets.outputs.targets_updated != 'true'
+      if: steps.update_targets.outputs.targets_updated != 'true' && steps.sign_targets.outputs.targets_signed != 'true'
       run: |
         if tuf-on-ci-status >> status-output;  then
           echo "status=success" >> $GITHUB_OUTPUT
@@ -143,19 +191,3 @@ runs:
           }
 
           await core.summary.addHeading(summary).write()
-
-    - name: Dispatch another signing event workflow
-      # dispatch if using default token: otherwise update_targets step
-      # has alreaddy triggered a push event
-      if: inputs.token == github.token && steps.update_targets.outputs.targets_updated == 'true'
-      uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
-      with:
-        github-token: ${{ inputs.token }}
-        script: |
-          console.log('Dispatching another signing event workflow after a targets metadata update')
-          github.rest.actions.createWorkflowDispatch({
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            workflow_id: 'signing-event.yml',
-            ref: process.env.GITHUB_REF_NAME,
-          })
--- a/actions/online-sign-targets/action.yml
+++ b/actions/online-sign-targets/action.yml
@@ -100,94 +100,18 @@ runs:
         cat sign-output >> "$GITHUB_STEP_SUMMARY"
       shell: bash
 
-    - id: status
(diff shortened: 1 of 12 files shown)


kommendorkapten:
e2e tests are currently failing. Will look into that.

signer/tuf_on_ci_sign/delegate.py
```diff
@@ -89,11 +92,35 @@ def verify_signers(response: str) -> list[str]:
...
+                        default=", ".join(config.signers),
+                        value_proc=verify_signers,
+                    )
+                elif signer_choice == 2:
+                    toplevel = git_expect(["rev-parse", "--show-toplevel"])
+                    settings_path = os.path.join(toplevel, ".tuf-on-ci-sign.ini")
+                    user_config = User(settings_path)
+                    online_key = _collect_online_key(user_config)
+                    uri = online_key.unrecognized_fields[TAG_ONLINE_URI]
+                    config.signers = [uri]
```
jku, line 123: I left some higher level comments on this in the issue: I'm not sure the online keys should be a special case -- why not consider them just one of the signers you can add?

repo/tuf_on_ci/online_sign.py
```diff
@@ -29,10 +29,13 @@ def _git(cmd: list[str]) -> subprocess.CompletedProcess:
 @click.command()  # type: ignore[arg-type]
 @click.option("-v", "--verbose", count=True, default=0)
 @click.option("--push/--no-push", default=False)
-def online_sign(verbose: int, push: bool) -> None:
+@click.option("--role", default="")
+def online_sign(verbose: int, push: bool, role: str) -> None:
```
jku, line 33: I left a higher level comment on the issue: it kind of looks like doing the signing in the repo code is correct but combining the current online-sign and the new feature make this function more complicated but does not have synergies: a new top-level CLI command might make sense

jku (commented):
Left some code comments in addition to https://github.com/theupdateframework/tuf-on-ci/issues/75#issuecomment-2257776296

signer/tuf_on_ci_sign/_signer_repository.py
```diff
@@ -760,9 +793,13 @@ def sign(self, rolename: str):
...
         for key in self._get_keys(rolename):
-            keyowner = key.unrecognized_fields["x-tuf-on-ci-keyowner"]
-            if keyowner == self.user.name:
+            # Detect if the key is an online key
+            if TAG_ONLINE_URI in key.unrecognized_fields:
                 signing_keys[key.keyid] = key
+            else:
+                keyowner = key.unrecognized_fields[TAG_KEYOWNER]
+                if keyowner == self.user.name:
+                    signing_keys[key.keyid] = key
```
jku, line 802: Is this necessary here? We're trying to sign online keys in repo tools not in signer

repo/tuf_on_ci/_repository.py
```diff
@@ -231,6 +241,35 @@ def close(self, rolename: str, md: Metadata) -> None:
 
         self._write(rolename, md)
 
+    def sign(self, rolename: str) -> None:
```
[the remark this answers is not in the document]
jku: actually this is not a relevant complaint: this design matches signer and makes sense:
* close() modifies the metadata (versions etc) and may sign as a "side effect"
* sign only signs without any modification

jku:
I have a completely untested branch (based on this one) with an action in https://github.com/jku/tuf-on-ci/tree/external-delegations

I'll have a look at tuf-on-ci-delegate next: I think we should be able to handle online signers just as "one of the signers" and not as a special case (IMO the special casing makes sense for timestamp/snapshot but not really for targets)

signer/tuf_on_ci_sign/delegate.py
jku: I guess my suggestion would require some re-design of adding signers since the KMS keys need more info so maybe not worth it

jku:
I have a seemingly working signing event in https://github.com/jku/test-online-targets/pull/3. Some notes:
* The decision to include in signing event seems good in that signing event is useful to see the status: this should work in external repo and upstream
* when used like this, the signing-event action of course needs similar things as online sign
  * token permission
  * kms account details (e.g. two new optional inputs for GCP)
  * the code to sign into kms
* we could still leave all of those out of the template workflow (and just document hos to add it): if the default signing-event does not need token permission then maybe we don't enable it by default ?

Overall I don't like how the most complicated bits of code got more complicated (in this PR as a whole): testing and refactoring is probably needed... but I think we can work with this

jku:
I've now tested both signing in the upstream project and signing externally. Both work, see the example event.

That said the external flow requires some polishing: 
* I modified workflows compared to upstream
   * I removed most workflows   
   * The signing-event workflow required changes (roles_to_sign and the KMS details)
* These changes mean that the signing event needs to be rebased on top of them when signing externally, and then the signature commit kind of needs to be cherry-picked to the real signing event branch -- because we wouldn't want the workflow modifications getting merged upstream

So the process works, it's just not very ergonomic. Not sure what would be the easiest workflow here

signer/tuf_on_ci_sign/delegate.py
```diff
@@ -40,14 +40,16 @@
...
+TAG_ONLINE_URI = "x-tuf-on-ci-online-uri"
 
 logger = logging.getLogger(__name__)
 
 
 def _get_offline_input(
     role: str,
     config: OfflineConfig,
-) -> OfflineConfig:
+) -> tuple[OfflineConfig, Key]:
```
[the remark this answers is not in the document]
jku: so I am wrong here: it is used (since otherwise we don't have the keyvalue) 

The original design is just bad and this makes it more visible. The key should likely just be a part of the OfflineConfig: `OfflineConfig.signers` is currently a `list[str]` and it probably should be a `list[tuple[str, Key | None]]` so each signer could have a key attached to them

kommendorkapten:
I've tested the updated changes now, and are happy with how they can be used. Will review the code now.

signer/tuf_on_ci_sign/delegate.py
```diff
@@ -89,11 +92,35 @@ def verify_signers(response: str) -> list[str]:
         if choice == 0:
             break
         if choice == 1:
-            config.signers = click.prompt(
-                bold(f"Please enter list of {role} signers"),
-                default=", ".join(config.signers),
-                value_proc=verify_signers,
-            )
+            # if role is not root or targets, allow online keys
```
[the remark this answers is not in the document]
kommendorkapten: The only reason I didn't let `targets` use online keys was to minimize the scope of this PR, i.e not change any existing behaviour. I'm fine to let `targets` use online keys though, I'll update this PR to support it.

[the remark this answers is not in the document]
kommendorkapten: This is now fixes.

signer/tuf_on_ci_sign/delegate.py
kommendorkapten: That was the reason I made them special, the code is easier. And also I think the flow for the repository maintainers is simplified if they just provide all the details for the online key direct.

signer/tuf_on_ci_sign/_signer_repository.py
kommendorkapten: It appears it's a left over from an earlier version, good catch. I tried locally with this hunk reverted and it works. Removing this.

repo/tuf_on_ci/signing_event.py
```diff
@@ -225,6 +225,67 @@ def update_targets(verbose: int, push: bool) -> None:
...
+
+    signed_targets = []
+    repo = CIRepository("metadata")
+    for role in roles:
+        if role in ["root", "timestamp", "snapshot"]:
+            raise ValueError(f"Non-targets role '{role}' cannot be signed")
+
+        if repo.is_signed(role):
+            click.echo(f"Signing not required for role '{role}'")
+            continue
```
jku, line 255: this kind of also encodes the assumption that there is a single signer: is_signed() returns true if role is signed by threshold of sigs (and is not in signing period) -- I suppose that's fine as long as online signing always means 1 signer

jku (commented):
One clearly missing thing is documentation: 
* We should mention how to setup online targets signing: maybe a new `docs/ONLINE_TARGETS_SIGNING.md` or a section in the existing online signing doc
* should document at least the needed changes in signing-event workflow
* I think the external signing part does not need to be explicitly documented
* maybe mention that this is still experimental? 

Otherwise I think we can live with this PR (as long as we file some issues and actually follow up later). I did leave some new code comments for discussion

jku:
btw, for the first time ever I'm now seeing `"You have exceeded a secondary rate limit"` in signing event when testing this (happens calling search api). 

I don't understand how this could be related to the PR but it is the first time I see this -- maybe this limit is per user and I just have enough projects now?

actions/signing-event/action.yml
```diff
@@ -20,6 +20,18 @@ inputs:
...
+    required: false
+    default: ""
+  gcp_workload_identity_provider:
+    description: "Google Cloud workload identity provider (required if GCP is used to sign targets roles)"
+    required: false
+    default: ""
+  gcp_service_account:
+    description: "Google Cloud service account name (required if GCP is used to sign targets roles)"
+    required: false
+    default: ""
```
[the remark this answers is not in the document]
kommendorkapten: Oh, good call out, I'll add Azure too.

actions/signing-event/action.yml
```diff
@@ -42,24 +62,41 @@ runs:
...
+        cat update-output >> output
+        cat update-output >> "$GITHUB_STEP_SUMMARY"
+      shell: bash
+
+    - id: sign_targets
+      if: inputs.targets_to_sign != ''
+      env:
+        TARGETS_TO_SIGN: ${{ inputs.targets_to_sign }}
+      run: |
+        if tuf-on-ci-online-sign-targets $TARGETS_TO_SIGN >> sign-output;  then
```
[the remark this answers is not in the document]
kommendorkapten: But that would be expected, right? First the updated metadata is pushed, then it's signed. Having them as separate commits makes sense. Are you referring to commit twice, but only push once? I don't think that's a big deal.

kommendorkapten:
> btw, for the first time ever I'm now seeing `"You have exceeded a secondary rate limit"` in signing event when testing this (happens calling search api).

I have been seeing this quite frequently, both on my personal test accounts and some real TUF repositories, so probably a coincidence that you started to see it now?

jku:
> I have been seeing this quite frequently, both on my personal test accounts and some real TUF repositories

Well, I filed an issue now that I saw it for the first time

jku:
This is something I was a bit afraid of when modifying signing-event (since it's self-dispatching):

https://github.com/jku/test-online-targets/pull/4

1. there is a signing-event for another role `externally-delegated-role`
1. because `delegated-role` is in signing period, signing-event ends up signing it... not sure we can easily prevent this
1. for some reason it keeps signing again and again -- as if it was constantly in signing period

* So there is the infinite loop to look at -- possibly it has to do with my the long signing period I used (signing period is just a day shorter than expiry)
* but also why is there no signing event for `delegated-role` itself if it is in signing period?

kommendorkapten:
> So there is the infinite loop to look at -- possibly it has to do with my the long signing period I used (signing period is just a day shorter than expiry) 

That is very worrying. I'll take a look.

kommendorkapten:
Somehow there is an event for `delegated-role`: https://github.com/jku/test-online-targets/compare/sign/delegated-role-v2

A bit unexpected but it's commented in the external delegated event: https://github.com/jku/test-online-targets/pull/4#issuecomment-2286970184

jku:
I think it may be just an issue in the date comparison for "are we in signing  period" (in `CIRepository.is_signed()`), I need to take a closer look.

I'll note that this would likely have been noticed if we had a e2e test...

jku:
> Somehow there is an event for `delegated-role`: https://github.com/jku/test-online-targets/compare/sign/delegated-role-v2

oh I know what that is about: the roles are "delegated-role" and "externally-delegated-role": somehow signing event for former finds the PR for the latter because of non-exact string matching

actions/online-sign-targets/action.yml
```diff
@@ -0,0 +1,193 @@
...
+          echo "targets_signed=true" >> $GITHUB_OUTPUT
+        else
+          echo "targets_signed=false" >> $GITHUB_OUTPUT
+        fi
+        cat sign-output
+        cat sign-output >> output
+        cat sign-output >> "$GITHUB_STEP_SUMMARY"
+      shell: bash
+
+    - id: status
```
jku, line 103: we should not have this step or the update-pr step here...
* when using custom token, signing event will trigger and do these steps
* when using default token, we will need to trigger signing-event from here: but that should be safe as it's not a loop then

jku:
### Summary of current approach and its issues 

* We're trying an approach where targets are signed in a new action that is called from a new workflow
* There are multiple ways this should get dispatched (in a normal single repository setup)
   * delegation changes done by maintainer: metadata is modified by user
   * targets files being changes by maintainer: metadata is modified by signing-event
   * create-signing-events bumping the versions: metadata is modified by create-signing-events
* if we use the default token the last two types of changes will not trigger a push event that the sign-online-targets workflow needs

I don't see a reasonable way to make the dispatching work in this design with default token (this is the way signing-event looks like it does otherwise too -- I'd like it to be more granular but it's not possible)


### Options

* Just ditch the action from the PR, admit that all of this code is purely for your use case for now and no-one else will use it: you maintain the action in your repo
* keep (roughly) the current PR. It will not work as is at least with default token (and I don't see a future path where it would)
* revert to signing in signing-event: this we can make work, just need to be careful to avoid dispatch loops

Not great but that's what it looks like :/

jku:
I'm also looking at the loop issue in the "signing integrated in signing-events" version:
* is_signed() is not the correct check in our case:
   * it makes sense in online-sign (timestamp/snapshots case) because then being in signing period actually means we want to sign
   * it does not make sense online-sign-targets  since the role _should_ not be in signing period because the expiry has already been updated: in fact being in signing period is a potential issue: if we sign now there would be a new signing event immediately after merge
   
I have fixed this in my branch and tested in the "infinite loop" signing event.

kommendorkapten:
It now works in this example: https://github.com/kommendorkapten/tuf-repo-test-4/pull/115

There are no loops. Note that I'm using the default token, so the delegation workflow is not [triggered](https://github.com/kommendorkapten/tuf-repo-test-4/blob/main/.github/workflows/new-delegate.yml#L7). I had to manually  approve that workflow to run.

A workaround for this is to not use the default token, but a custom PAT, that would trigger this workflow to run.

I'm fine with this, given that the only _current_ usecase for this is the npm delegation for Sigstore. That will be manged in a separate repo, so the signing workflow will be triggered upon a change, as the push will not happen via the default gh token.

If we accept this, we should create an issue to track this, as for future scenarios, having the complete signing event working is desirable as it would unlock more use-cases for tuf-on-ci.

jku:
Added a commit that makes `is_signed()` actually do what the name says: we are not interested in signing period in the targets signing case (since that is done by create-signing-events)

jku (approved):

jku:
We can go with this, let's see how it works

PR #409, opened 2024-08-22
Add a temporary hack to always generate sha512 prefixed artifacts

jku:
This is done even if sha512 is not used in the metadata:
* sigstore-java assumes this prefix is used
* this temporary fix means sigstore root-signing does not need to break a client without warning

--- 

Leaving this a draft until I've tested


jku:
I've tested, seems to work. I think we could consider:
* merging this, release a 0.12.1
* deploying that in root-signing-staging, re-running online-sign: this should make the test during publish pass

jku:
making draft in light of current discussion in https://github.com/sigstore/root-signing-staging/issues/161: maybe we don't need this

PR #416, opened 2024-08-27
signer: Accept username without @ in config file

jku:
It seems harmless but useful to prefix usernames with @ if the username in configuration does not have the prefix already


joshuagl (approved):
Thanks for addressing this so quickly.

Change and test looks good!

PR #423, opened 2024-08-31
repo: Add workaround for sigstore KMS keyid

### LINKED ISSUE #422: investigate workaround for root-signing issue
https://github.com/sigstore/root-signing/issues/1347 is an issue with the root-signing migration signing event
* keyid for GCP KMS key is incorrect
* the deadline for finishing the migration is getting closer
* proper fix would require the keyholders to resign before we can re-test

Investigate whether we could add a special case workaround for this specific keyid: this might  allow the signing event to move forward -- the keyid could then be fixed in next signing event

jku:
The sigstore root-signing online key keyid was entered incorrectly: Add a workaround here so there is more time to fix the actual keyid.

Fixes #422

--- 
DRAFT while I am not sure of the actual key version

### DIFF
--- a/repo/tuf_on_ci/_repository.py
+++ b/repo/tuf_on_ci/_repository.py
@@ -228,6 +228,15 @@ class CIRepository(Repository):
         for key in self._get_keys(rolename):
             if rolename in ["timestamp", "snapshot"]:
                 uri = key.unrecognized_fields[TAG_ONLINE_URI]
+
+                # FIXME: workaround for issue #422, only needed while sigstore
+                # root-signing online key keyid is incorrect
+                if (
+                    uri
+                    == "gcpkms://projects/sigstore-root-signing/locations/global/keyRings/root/cryptoKeys/timestamp"
+                ):
+                    uri = "gcpkms:projects/sigstore-root-signing/locations/global/keyRings/root/cryptoKeys/timestamp/cryptoKeyVersions/1"  # noqa: E501
+
                 signer = Signer.from_priv_key_uri(uri, key)
                 md.sign(signer, True)
             else:


jku:
I don't have the GCP permissions to test this but Bob confirmed the current version is 1.

jku:
There were actually two issues with the key, both appeared because I just copied the similar looking but not same format from legacy root-signing:
* should start with just "gcpkms:" not "gcpkms://"
* must contain version number in the end

This is the incorrect legacy format:
`gcpkms://projects/sigstore-root-signing/locations/global/keyRings/root/cryptoKeys/timestamp`
this is the fixed format:
`gcpkms:projects/sigstore-root-signing/locations/global/keyRings/root/cryptoKeys/timestamp/cryptoKeyVersions/1`
here is another key that works on tuf-on-ci for comparison:
`gcpkms:projects/python-tuf-kms/locations/global/keyRings/tuf-demo/cryptoKeys/snapshot/cryptoKeyVersions/1`

repo/tuf_on_ci/_repository.py
```diff
@@ -228,6 +228,15 @@ def close(self, rolename: str, md: Metadata) -> None:
         for key in self._get_keys(rolename):
             if rolename in ["timestamp", "snapshot"]:
                 uri = key.unrecognized_fields[TAG_ONLINE_URI]
+
+                # FIXME: workaround for issue #422, only needed while sigstore
+                # root-signing online key keyid is incorrect
```
kommendorkapten, line 233: Instead of this hard coded rewrite, how about a version that detects invalid format (wrong scheme and missing version) and use that before instantiating the key via `Signer.priv_key_from_uri`?

kommendorkapten (approved):

Issue #399, opened 2024-08-13
Use rest.pulls API instead of rest.search to fix timeouts


jku:
signing-event uses:

          const prs = await github.rest.search.issuesAndPullRequests({
            q: `in:title+"${title}"+state:open+type:pr+repo:${repo}`
          })

to find the signing event PR.

The search API has very low usage limits. We should try using the pulls API:

```
github.rest.pulls.list({
  owner=owner,
  repo=repo,
  state="open",
  per_page=100
});
```
and finding the matching PR by matching the title

jku:
related issue: the `in:title` search is not an exact match: we should fix that so partly matching role names don't create false positives

PR #469, opened 2024-11-06
Avoid using search, look up pr by current commit instead

kommendorkapten:
Fixes: https://github.com/theupdateframework/tuf-on-ci/issues/399

The assumption is that this endpoint is way (computationally) cheaper then the org/repo search endpoint.

During the list PR we provide the tip of the current branch, during signing events is very unlikely that the tip is part of another PR, so I would think the code 'as is' is safe for use.

Marking this as draft for now, as I want to run more tests to confirm that this does not trigger the secondary rate limit.

### DIFF
--- a/actions/signing-event/action.yml
+++ b/actions/signing-event/action.yml
@@ -74,14 +74,15 @@ runs:
         script: |
           const fs = require('fs')
           const title = `Signing event: ${process.env.GITHUB_REF_NAME}`
-          const repo = `${context.repo.owner}/${context.repo.repo}`
-          const prs = await github.rest.search.issuesAndPullRequests({
-            q: `in:title+"${title}"+state:open+type:pr+repo:${repo}`
+          const prs = await github.rest.repos.listPullRequestsAssociatedWithCommit({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                commit_sha: context.sha
           })
 
-          if (prs.data.total_count > 1) {
-            core.setFailed("Found more than one open pull request with same title")
-          } else if (prs.data.total_count == 0) {
+          if (prs.data.length > 1) {
+            core.setFailed("Found more than one open pull request the current commit")
+          } else if (prs.data.length == 0) {
             const response = await github.rest.pulls.create({
               owner: context.repo.owner,
               repo: context.repo.repo,
@@ -94,7 +95,7 @@ runs:
             pr = response.data.number
             console.log(`Created pull request #${pr}`)
           } else {
-            pr = prs.data.items[0].number
+            pr = prs.data[0].number
             console.log(`Found existing pull request #${pr}`)
           }
 


jku (approved):
Yeah this should avoid the search rate limit (30/minute) and use the core rate limit instead (5000/hour) which is likely to be better in most situations.

According to docs the branch name would work instead of sha as well... but the current implementation seems fine to me. Let's go with this.

kommendorkapten:
I ran six concurrent signing events, and the rate limit didn't trigger (it's usually triggered for a single signing event), so I'll take his from draft to ready for review as it seems to work as intended.

jku:
:+1: 

> it's usually triggered for a single signing event

documenting for posterity: I don't know the repo/project in question but my assumption is that the same user/app token is running other workflows as well and combined these workflows (ab)use the search API with > 30 queries per PR. The tuf-on-ci workflows alone do not trigger this rate limit -- in fact I've never seen this issue with tuf-on-ci yet.

kommendorkapten:
Yes, that's a good call-out, we have some other workflows that are running too (but not accessing the search API).

jku:
Well that's interesting because then I don't understand how this could help.... 

If you can please get some rate limit quota measurements -- would have been great to get those before the PR as well

kommendorkapten:
So we have this events (from October, last signing ceremony):
1. Create signing event runs (create the branch)
2. Run signing event, fails with the following:

```
      date: 'Tue, 15 Oct 2024 07:24:21 GMT',
      'x-ratelimit-limit': '30',
      'x-ratelimit-remaining': '30',
      'x-ratelimit-reset': '1728977121', (Tis 15 Okt 2024 07:25:21 UTC)
      'x-ratelimit-resource': 'search',
      'x-ratelimit-used': '1',
```

See the previous runs here:
Screenshot 2024-11-07 at 12 44 31

So I only see the rate limit quota for failed requests, but none of the preceding runs should exhaust it. I'm not sure about the details for calculating the estimated cost for a search query, not sure it's public for secondary rate limits: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits

kommendorkapten:
I managed to trigger the secondary rate limit on my personal vanilla repo too just now: https://github.com/kommendorkapten/tuf-repo-test-4/actions/runs/11722913337

The previous action ran 1h43min before the one that failed, which should indicate that rate limit bucket for `search` is fresh. _Unless_ it's shared with my other account activity, but I'm not sure about the details.

jku:
> The previous action ran 1h43min before the one that failed, which should indicate that rate limit bucket for `search` is fresh. _Unless_ it's shared with my other account activity, but I'm not sure about the details. 

Based on this and previous comment (thanks for looking into it) this has nothing to do with the documented rate limit restrictions... I would guess it's account related since I don't remember ever seeing it

PR #485, opened 2024-12-04
TUF upgrade

jku:
This is a TUF, securesystemslib (and by extension sigstore) dependency upgrade
* Both TUF and securesystemslib had some significant work done in these releases: need to test well before deploying
* Tried to keep changes to minimum, so some new TUF API is not used on purpose to achieve that


jku:
I've read it through and tested locally: Seems good to go to me

kommendorkapten (approved):

PR #504, opened 2025-01-10
actions: Switch status html rendering to pandoc

### LINKED ISSUE #430: repository status page styles break when repo is moved
compare
https://sigstore.github.io/root-signing/index.html
vs
https://tuf-repo-cdn.sigstore.dev/index.html

This is because 
* tuf-on-ci only produces the index.md, the html is then produced from markdown with jekyll
* jekyll really, really likes using absolute URLs so the css URL is `/root-signing/assets/css/style.css?v=a4b685e8705fdb5078638a18dbea969788c0036d`

when the repository is moved even if we also move the stylesheet, this absolute path is incorrect.

Possible fixes:
* get jekyll to use relative paths -- this seems impossible
* hack the html after it is generated -- this is not trivial since the the html is in metadata dir whereas the stylesheet is not...
* produce the html in tuf-on-ci directly -- this is likely best alternative but I don't have the skills
* use something else than jekyll -- I chose jekyll-build-pages only because it's a GitHub maintained action).

I think the last option might be worth looking into: we could e.g. make `tuf-on-ci-build-repository` depend on python-markdown and do the conversion there, or use the `docker://pandoc/core:2.9` action

jku:
We used jekyll before: it has various issues and turned out to not be the right tool for the job -- I only used it since it was a GitHub maintained action.

* Use pandoc (as container step) instead
* Include a minimal CSS file to make the status page look like it's not still the 1980s
* Prepare by copying the css file to workdir (which the pandoc container has access to)
* Two separate build directories are now not needed (everything happens in "build/").
* The result is now full standalone: copying the html file around should break nothing 

Fixes #430.

Example in https://jku.github.io/tuf-demo/metadata/

### DIFF
--- a/actions/upload-repository/action.yml
+++ b/actions/upload-repository/action.yml
@@ -47,7 +47,7 @@ runs:
       shell: bash
 
     - name: Render repository state in HTML
-      uses: docker://pandoc/core:3.5
+      uses: docker://pandoc/core:3.5.0@sha256:771842ce6f661785e0e12931f82ea64046d70fa48ea5cff480492d79ab3b8ff1
       with:
         args: >-
           --metadata title="TUF Repository state"
--- a/actions/upload-repository/action.yml
+++ b/actions/upload-repository/action.yml
@@ -34,6 +34,7 @@ runs:
         echo "::group::Install tuf-on-ci"
         ROOT=$GITHUB_ACTION_PATH/../..
         pip install -c $ROOT/action-constraints.txt $ROOT/repo/
+        cp $GITHUB_ACTION_PATH/index.css .
         echo "::endgroup::"
       shell: bash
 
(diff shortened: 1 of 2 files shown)


actions/upload-repository/action.yml
```diff
@@ -45,27 +46,34 @@ runs:
...
 
-    - uses: actions/jekyll-build-pages@44a6e6beabd48582f863aeeb6cb2151cc1716697 # 1.0.13
+    - name: Render repository state in HTML
+      uses: docker://pandoc/core:3.5
       with:
-        source: build
-        destination: build-jekyll
+        args: >-
+          --metadata title="TUF Repository state"
+          --variable title=""
```
jku, line 54: this weirdness is required so
* the page (tab) has a title and pandoc does not complain
* but the title is not used as a h1 (since we have a header already)

actions/upload-repository/action.yml
```diff
@@ -45,27 +46,34 @@ runs:
         find build -type f | xargs ls -lh
       shell: bash
 
-    - uses: actions/jekyll-build-pages@44a6e6beabd48582f863aeeb6cb2151cc1716697 # 1.0.13
+    - name: Render repository state in HTML
+      uses: docker://pandoc/core:3.5
```
kommendorkapten, line 50: Should we pin by hash here?

kommendorkapten (approved):
Looks good overall, only one comment about pin to hash.

actions/upload-repository/action.yml
jku: yeah maybe that is better -- it does not look like a fast moving target anyway

PR #513, opened 2025-01-15
repo: Do not use naive datetimes with tuf API

### LINKED ISSUE #512: Use of naive datetime
The test client for repository (`repo/tuf-on-ci/client.py`) uses a tweaked tuf client to  verify that repository is valid X days into future: this hack still seems to use naive datetimes which now broke in https://github.com/sigstore/root-signing-staging/issues/210 because python-tuf client uses non-naive datatimes now

Fix should be easy, we can make a patch release after that

jku:
The datetime format is not really part of TUF API but we meddle in in the client internals to do the "valid X days from now" test... so need to match the used datetime format.

Fixes #512 

I've tested this in https://github.com/jku/tuf-demo/pull/122: seems to work ok -- I'll also modify tuf-demo so that we see it work after merge


jku:
e2e tests could probably cover this... but I'd like to get this fix released soon: I'm marking this draft so I can include the version bump right here.

kommendorkapten:
Looks good. Do you know what was the breaking change that introduced this?

kommendorkapten (approved):

jku:
> Looks good. Do you know what was the breaking change that introduced this?

Yes: the python-tuf API did not change ubt tuf-on-ci-test-client script (used in test-repository actions) tweaks one thing python-tuf client internals... that one thing is expiry which now needs to be in UTC not "naive" datetime.

PR #514, opened 2025-01-16
Bump repository and test Python version to 3.13

jku:
This is not urgent in any way but good to do at some point (and doing it right after a release sounds like a good idea):
* Install Python 3.13 in all actions
* Use 3.13 in tests and other workflows (except signer tests where we also want 3.9)


kommendorkapten (approved):
Looks good.

PR #520, opened 2025-01-21
Prepare for release v0.15.2

kommendorkapten:


jku:
the only possible issue I have is that #514 was merged. So if we make a release from main that will be included -- should not break anything but I still don't think it should go into a point release.

So I'd suggest one of two things:
* revert #514 then release from main
* make a 0.15 branch, release from there

It looks like the release workflow has no special checks so the second option probably just works: I can try that

jku:
reverted the upgrade for now, I will now merge this

jku (approved):

PR #528, opened 2025-01-24
signer: Handle the "yubikey auth required" case

### LINKED ISSUE #527: Improve error message if touch is not provided
```
Failed to sign targets with @-username- key.
  Failed to sign: CKR_USER_NOT_LOGGED_IN (0x00000101)
```

Currently output is the above if user does not realize their key requires a touch. We could make the message better. I think we can't easily tell what kind of authentication is required but _"Signing failed because a required authentication (e.g. touch) was not done"_ might work.

CC @mnm678

jku:
There is now light infrastructure for handling PKCS errors... I'm only handling the one that seems to trip people up since there are so many possible errors: we can't possibly do a good job of handling all of them.

Couple of review notes:
* I can't directly test this one: The touch policy on a yubikey can only be set when the certificate is created and and the touch policy on my keys is the default "never" -- I'd have to overwrite my key or buy a new yubikey for this. I have tested the code by adding handling for another case (CKR_ARGUMENTS_BAD) and that works fine so I trust that this one works too
* I was a little surprised I have to access a private field (`e.__context__`) to get the original exception... but that seems the only way

Fixes #527

### DIFF
--- a/signer/pyproject.toml
+++ b/signer/pyproject.toml
@@ -36,6 +36,7 @@ python_version = "3.9"
 [[tool.mypy.overrides]]
 module = [
   "securesystemslib.*",
+  "PyKCS11.*",
 ]
 ignore_missing_imports = "True"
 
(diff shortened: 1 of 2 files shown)


kommendorkapten (approved):
Code looks good. I don't have a yubi with touch policy either, co can not test.

PR #536, opened 2025-02-03
repo: Ensure keyids are correct in signing event

jku:
Keyids should match the hash of the key content. The signing tool already does this but makes sense to verify in signing-event (since metadata can be modified outside the signing tools).

Upstream python-tuf does not enforce keyids since I/ we think that
* this provides no additional security and
* forcing keyid updates is not a good idea and makes repositories a pain to maintain

This is all true but because other clients (awslabs/tough) disagree, this now becomes an interop issue 

See https://github.com/sigstore/root-signing/issues/1431


jku:
I've added a release prep commit (with changelog and version number changes), with the assumption we want to do a release.

PR #543, opened 2025-02-05
Bump repository and test Python version to 3.13

### LINKED ISSUE #523: unrevert #514
We reverted so we could release another point release from main.

There's no rush with this one but let's remember to un-revert at some point

jku:
This is a second attempt to merge this: it was reverted since we needed to do a quick point release from main.

* Install Python 3.13 in all actions
* Use 3.13 in tests and other workflows (except signer tests where we also want 3.9)

Fixes #523


jku:
merging this now, I want to do some PRs on top of this

PR #770, opened 2026-04-16
[ARTSEC-INT] kipz/private repo support

harmonherring-pro:
Brings in kipz's changes for private repo support


kommendorkapten:
👋 Can please share more details around the context for this? It's totally possible today to run tuf-on-ci with a private repository so I'm not really sure what this is supposed to solve.

kommendorkapten:
Is the problem that you also have the GitHub page visibility set to private?

Issue #771, opened 2026-04-16
Instructions for initializing Yubikey PIV use unsupported tool. Update to use CLI.


elmatterino:
According to https://www.yubico.com/support/download/yubikey-manager/, the "Yubikey Manager" GUI tool used by instructions docs/YUBIKEY-PIV-SETUP.md is now out of support. The Yubikey Manager CLI is recommended by Yubikey.

PR #772, opened 2026-04-16
Update YubiKey PIV instructions to ykman CLI

elmatterino:
Update the YubiKey PIV initialization instructions to use the ykman CLI since the Yubikey Manager UI is no longer supported by Yubico.

Fixes issue #771

### DIFF
--- a/docs/YUBIKEY-PIV-SETUP.md
+++ b/docs/YUBIKEY-PIV-SETUP.md
@@ -1,14 +1,17 @@
 # YubiKey PIV Setup
 
-This guide walks through using the YubiKey Manager UI to configure PIV for TUF signing operations.
+This guide walks through using the YubiKey Manager CLI to configure PIV for TUF signing operations.
 
 > [!IMPORTANT]
 > This is a general setup guide for using YubiKey with TUF-on-CI. When provisioning YubiKeys for production TUF use, you may want to consider additional procedures around the procurement, distribution, and configuration of the devices. (e.g. serial number tracking, offline device configuration, YubiKey Manager CLI, hardware random number generators, etc.)
 
 ### Requirements
 
-- Download [YubiKey manager](https://yubico.com/support/download/yubikey-manager/):
-  ![Yubikey manager UI](yubikey-manager.png)
+Download YubiKey Manager CLI (ykman)
+
+```shell
+pip install yubikey-manager
+```
 
 > [!TIP]
 > Use https://www.yubico.com/genuine/ to confirm that your YubiKey device is genuine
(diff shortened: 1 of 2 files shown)


docs/YUBIKEY-PIV-SETUP.md
```diff
@@ -26,55 +29,52 @@ PIN codes are used for signing operations and to unlock a device.
...
+
+```shell
+ykman piv keys generate -a eccp256 9c pub.pem
+ykman piv certificates generate 9c --subject subject-field-value pub.pem
+```
+
+It is recommended to use your GitHub handle as the `--subject` parameter value.
+
+> [!NOTE]
+> The tuf-on-ci-delegate sign/init command currently requires the key type to be eccp256.
```
jku, line 78: can you explain this? Is this some pkcs11lib limitation?

docs/YUBIKEY-PIV-SETUP.md
```diff
@@ -26,55 +29,52 @@ PIN codes are used for signing operations and to unlock a device.
...
-1. Select `Self-signed certificate` unless you are using a CA to issue the certificate via a CSR
-1. Select the algorithm to be used for digital signatures
-1. Populate the `Subject` field, this can be anything (recommend using your GitHub handle)
-1. Select an `Expiration date`, this is the date the signing cert expires but the signing keys can still be used for TUF-on-CI after expiration
-1. Select `Generate`
+Generate a key in the signature slot (PIV slot 9c). The following generates a self-signed certificate.
+
+```shell
+ykman piv keys generate -a eccp256 9c pub.pem
+ykman piv certificates generate 9c --subject subject-field-value pub.pem
```
jku, line 72: the default lifetime is just one year... but maybe we don't need to complicate this  with `--valid-days` since I don't think expired certs are a problem here (I believe yubikeys will happily sign with expired certs). So I guess this is fine

jku (commented):
I think the Authenticator GUI app can still do all this as well... but I think I'm happy to take this PR -- it's probably easier to document this than a different GUI.

Just checking: you have tested these commands, right?

elmatterino:
I kept a history of commands I used to initialize a Yubikey 5C to work with a tuf-on-ci deployment example and derived these instructions directly from that history, so they have been tested.

docs/YUBIKEY-PIV-SETUP.md
elmatterino: I originally generated the PIV key as RSA 2048 and the tuf-on-ci-delegate sign/init command failed with a message that only ECC keys are supported. I regenerated the PIV key as eccp256 and the process succeeded. The comment reflects that discovery.

docs/YUBIKEY-PIV-SETUP.md
elmatterino: There's a comment in the original instructions (original line 77) about the key still being usable for TUF-on-CI after the cert expiration date. It's why I didn't worry about the --valid-days argument in the case of a self-signed certificate. The Yubikey isn't actually interpreting the cert. It's up to the PIV consumer to make that sort of decision, and TUF-on-CI isn't using it.

docs/YUBIKEY-PIV-SETUP.md
jku: Thanks. I just checked and it is a securesystemslib limitation: 
* only ECDSA (256, 384) is currently supported
* RSA is possible, likely a small change
* ED25519 is also theoretically possible (but in practice is maybe not useful)

jku (approved):

jku:
Thanks!

PR #792, opened 2026-05-25
Added support to disable key id verification.

kommendorkapten:
Disablement happens by setting the environment variable TUF_ON_CI_NO_VERIFY_KID to a non empty value.

See https://github.com/kommendorkapten/tuf-repo-test-3/pull/52 for a test event.

First we get a warning of key id mismatch, then the event action is updated to this branch, and no warnings are printed when retriggered.

### DIFF
--- a/repo/test/test_ci_repository.py
+++ b/repo/test/test_ci_repository.py
@@ -1,10 +1,44 @@
 import os
 import shutil
 import unittest
+from copy import deepcopy
 from tempfile import TemporaryDirectory
+from unittest import mock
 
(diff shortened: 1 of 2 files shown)


jku (approved):
lgtm.

For reference to future readers: this happens because 
* the spec does say that keyid must match the content
* but requiring that is not a security improvement (in fact it's arguably opposite as readable keyids would lead to better security)
* it's very easy to make a mistake in the keyid calculation as shown here 
* while most clients do not enforce keyid matching, awslabs/tough does

PR #802, opened 2026-06-04
build: Let's try uv

### LINKED ISSUE #614: try dependabot for pinned action-constraints.txt
I've seen it mentioned that dependabot can actually use pip-compile in some situations: this would be preferable to the hack workflow we now have...

I assume there is some heuristic in dependabot  where a `<FILE>.in` and a `<FILE>.txt` in same directory trigger using pip-compile... and our case hasn't worked because there is no .in file currently: should test this

jku:
This is probably my 4th attempt at usable dependency hash pinning

* Use uv instead of pip, uv_build instead of hatchling 
* Use uv.lock -- actions now pin based on uv.lock, and so does tox for the test envs
* Even uv itself is now pinned (setup-uv will lookup the version in uv.lock)
* Drop tox, use uv environments directly from Makefile instead
* dependabot should now be able to bump uv.lock (but you can't make dependabot update transient deps if they are still compatible :facepalm: ). So update-pinned-deps.yml remains but now uses `uv lock upgrade` -- I'm willing to discuss this: I'm not sure how important frequent transient dep upgrades really are...
* Remove actions-lint.yml (now handled as part of normal ci matrix)

This is going to need a bit of testing... I'm currently running https://github.com/jku/tuf-on-ci-sigstore-test on this branch, it seems fine


jku:
Just to make it clear: the tox test environments (locally and on CI) are still using pip:
* We could make them use uv as well (via tox-uv)
* or we could try replacing tox with a slightly more complicate Makefile and just uv 

but I wanted to keep this PR semi reasonable

jku:
another not-implemented feature: I have not pinned uv itself... because it seems very hard to do in a way that dependabot could handle :facepalm: 

This does look promising although it's not exactly clear: https://github.com/astral-sh/setup-uv/blob/main/docs/advanced-version-configuration.md#install-a-version-defined-in-a-requirements-or-config-file. Our case is also trickier because the config file is going to be in  $GITHUB_ACTION_PATH...

jku:
I did figure out a way to pin uv -- it's a bit clumsy but works

jku:
I think this an improvement

docs/DEVELOPMENT.md
```diff
@@ -1,33 +1,31 @@
...
 
 ```shell
 # Clone the project
 git clone https://github.com/theupdateframework/tuf-on-ci.git
 cd tuf-on-ci
-# Create virtual environment
-python3 -m venv .venv
-# Enter environment
+# Create virtual environment and install the signing and repository tools
+uv sync
```
jku: could includ `--frozen` here so people don't get uv.lock changes

signer/pyproject.toml
```diff
@@ -27,11 +27,8 @@ tuf-on-ci-delegate = "tuf_on_ci_sign:delegate"
 tuf-on-ci-import-repo = "tuf_on_ci_sign:import_repo"
 tuf-on-ci-sign = "tuf_on_ci_sign:sign"
 
-[project.optional-dependencies]
-lint = [
-  "mypy == 1.19.1",
-  "ruff == 0.15.14",
-]
+[dependency-groups]
+lint = ["mypy", "ruff"]
```
jku: since we have uv.lock, we no longer need to pin versions here

Makefile
```diff
@@ -0,0 +1,25 @@
...
+	tox -e test-signer
+
+test-repo: action-constraints.txt
+	tox -e test-repo
+
+test-e2e: action-constraints.txt
+	tox -e test-e2e
+
+action-constraints.txt: uv.lock
+	cd repo && uv export --quiet --no-emit-workspace --no-hashes -o ../action-constraints.txt
```
jku: this is here since I did not attempt to uv-ify tox yet: so when tests are run tox still runs pip and needs the action-constraints file, but it is generated from uv.lock now and used only in tests.

jku:
@kommendorkapten no rush with this one but when you do have a look:
* please give the local dev setup a test run: I had to change it a bit but it feels pretty good to me -- e.g. `make lint` just works in an empty source checkout without any prep (as long as you have `uv`). You can still enter the venv as usual to do anything more manually
* https://github.com/jku/tuf-on-ci-sigstore-test/ runs on this branch

.github/workflows/actions-lint.yml
```diff
@@ -24,15 +24,15 @@ jobs:
...
         python-version: '3.14'
-        cache: 'pip'
-        cache-dependency-path: |
+        enable-cache: true
+        cache-dependency-glob: |
           actions/lint-requirements.txt
 
     - name: Install zizmor
-      run: python -m pip install -r actions/lint-requirements.txt
+      run: uv tool install zizmor -c actions/lint-requirements.txt
```
jku: I think we can add zizmor to lint/dev groups and get the pinning without the extra file

actions/online-sign-targets/action.yml
```diff
@@ -75,14 +75,17 @@ runs:
...
+        version-file: ${{ github.action_path }}/../../repo/pyproject.toml
 
-    - run: |
+    - run: | # zizmor: ignore[github-env]
         echo "::group::Install tuf-on-ci"
-        ROOT=$GITHUB_ACTION_PATH/../..
-        pip install -c $ROOT/action-constraints.txt $ROOT/repo/
+        cd $GITHUB_ACTION_PATH/../../repo
+        UV_PROJECT_ENVIRONMENT=$GITHUB_WORKSPACE/.venv uv sync --frozen
+        echo "$GITHUB_WORKSPACE/.venv/bin" >> $GITHUB_PATH
```
jku, line 88: we could avoid this PATH change by either
* entering the venv in the later steps or
* calling the binaries with path (`./venv/bin/tuf-on-ci-online-sign ...`)

actions/create-signing-events/action.yml
```diff
@@ -15,14 +15,18 @@ runs:
...
         python-version: "3.14"
+        version-file: ${{ github.action_path }}/../../repo/pyproject.toml
 
-    - run: |
+    - run: | # zizmor: ignore[github-env]
         echo "::group::Install tuf-on-ci"
         ROOT=$GITHUB_ACTION_PATH/../..
-        pip install -c $ROOT/action-constraints.txt $ROOT/repo/
+        cd $ROOT/repo
+        UV_PROJECT_ENVIRONMENT=$GITHUB_WORKSPACE/.venv uv sync --frozen
```
jku: this should likely include `--no-dev`: I believe the default is to include the dev group as well

jku:
I added this in the description but raising this up for review:
**Transient deps in uv.lock are not upgraded greedily**: Unliuke direct deps, they only get upgraded as security upgrades or if compatibility breaks. So over time transient deps get stale.

Long term this sounds like a bad way to maintain a python project. Maybe we should still have a workflow that runs `uv lock --upgrade` once a month.

I'll mark this draft and try this

jku:
Ok, maybe this is now finally ready for review. It ballooned up a little since everything kind of depends on everything else

.github/workflows/ci.yml
```diff
@@ -38,24 +41,26 @@ jobs:
...
-        cache: 'pip'
-        cache-dependency-path: |
+        enable-cache: true
+        version-file: pyproject.toml
+        cache-dependency-glob: |
           signer/pyproject.toml
           repo/pyproject.toml
-          action-constraints.txt
-          build/build-constraints.txt
+          repo/uv.lock
```
jku: this should be just uv.lock

kommendorkapten:
@jku I'll try this out next week 👍

jku:
> @jku I'll try this out next week 👍

cheers. The docs maybe don't make this 100% clear but the pattern is:
* `uv` binary is now a required dev tool (like python) that needs to be in the environment -- this is annoying but manageable: personally I've created a `~/.tool-venv/` where I installed uv, and then I added `$HOME/.tool-venv/bin` to PATH in my ~/.profile
* All make targets should just work without manually creating or entering any venvs
* you can enter the dev env for manual testing though (`source .venv/bin/activate`)
* alternatively you can run things with `uv run`
  * what tools are available depends on which directory you are in: in git repo root dir, all exectuables and tools are available (so `uv run tuf-on-ci-test-client` or `uv run mypy .` all work)
  * in signer/ the repo executables are not available and vice versa)

.github/workflows/release.yml
```diff
@@ -17,16 +17,15 @@ jobs:
...
           python-version: '3.14'
-
-      - name: Install build dependencies
-        run: python3 -m pip install -c build/build-constraints.txt build
+          version-file: pyproject.toml
 
       - name: Build release changelog, signer wheel & source tarball
         run: |
-          PIP_CONSTRAINT=build/build-constraints.txt python3 -m build --sdist --wheel --outdir dist/ signer/
+          uv build --sdist --wheel --out-dir dist/ signer/ -b build/build-constraints.txt
```
jku: as long as we use uv_build we don't need build constraints (as uv_build comes from the uv being used)

actions/create-signing-events/action.yml
```diff
@@ -15,14 +15,17 @@ runs:
         fetch-depth: 0
         persist-credentials: true
 
-    - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+    - uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
       with:
+        enable-cache: false
         python-version: "3.14"
+        version-file: ${{ github.action_path }}/../../pyproject.toml
```
jku: uv has just released a version that supports uv.lock as version-file so this weird locking mechanism isn't needed

jku:
I did end up removing tox completely, this makes the whole thing a little easier to understand (Makefile now directly uses uv to run things in separate virtualenvs).

I'm going to merge this soon: Maintaining the branch for long term is too much work. If it does not work, we revert

jku:
CI is borken in the whole theupdateframework org: ubuntu-latest is never available

PR #815, opened 2026-06-29
signer: Support Tillitis TKey

jku:
* Enable ML-DSA use in general
* Start storing priv key uris in user data dir (on linux  ~/.local/share/tuf-on-ci-sign/signing-keys.ini) automatically: this is not just for tkeys but for all keys
* Add support for TKey as signing method
* Update yubikey code as needed for the pkcs11 changes in securesystemslib

The priv key uri caching was finally implemented because TKey really needs the uri for long term use: There is only one device binary for now but that's not going to be the case forever -- and switching device binaries changes the key content so the device binary must be encoded in the private key uri.

### Storing priv key uris in user data

There's a compromise here: We're avoiding putting the key details in metadata (like we do for online keys) because
  * It's information we don't need to share with world
  * Modifying metadata is not easy 
  
  The downside of this compromise is that
  * keyholder signing on a new device (laptop) is now trickier as the key details are in the application data file on the old laptop: I've added basic recovery for tkeys but it will only work as long as device binary remains same 
  
  https://jku.github.io/pqc-test-repo2/metadata/


jku:
A review note for later: during the initial key import and first signing, because of the peculiar design we can't check if there's e.g. a typo in the passphrase: the user would have to unplug the device between initial import_() and first signing for that to happen...
* we could ask for the passphrase twice
* or we could just say immediately after import "unplug the key and press enter..."

_EDIT_: I did the second option

jku:
This is now failing because I updated the securesystemslib reference to main branch:
main now has tkey support but it also has the pkcs library change, and apparently we have PyKCS11 code in this repo...

jku:
I will switch this to write to a separate state file for signing uris:
* moving to a new device is now potentially more difficult (because the state file that contains signer uris is not really user visible)
* but overwriting the users config file just felt wrong
* we should be able to recover in most situations, and putting keys in the config file still works -- so if you know what you're doing you can recover anything

signer/src/tuf_on_ci_sign/_signer_repository.py
```diff
@@ -280,19 +282,25 @@ def _get_keys(self, role: str, known_good: bool = False) -> list[Key]:
...
     def _sign(self, role: str, md: Metadata, key: Key) -> None:
         while True:
-            signer = self.user.get_signer(key)
             try:
+                signer = self.user.get_signer(key)
                 sig = md.sign(signer, True)
                 key.verify_signature(sig, md.signed_bytes)
                 self.user.set_signer(key, signer)
                 break
+            except RuntimeError:
```
jku: this is now KeyMismatchError

jku:
I think I'll do a securesystemslib release on monday and then merge this one: then we can do some real testing on all of the recent tuf-on-ci changes

signer/src/tuf_on_ci_sign/_user.py
```diff
@@ -48,11 +50,18 @@ def __init__(self, path: str):
                 f"Failed to find required setting {e} in {path}"
             ) from e
 
-        # signing key config is not required
-        if "signing-keys" in self._config:
-            self._signing_key_uris = dict(self._config.items("signing-keys"))
+        # User cache/state for signing keys
+        if state_path is None:
+            state_dir = user_state_dir("tuf-on-ci-sign")
```
jku: I think we could just use `user_data_dir`:
* it's more common and seems to fit our use fine
* only changes things on linux, other platforms use the same dir anyway

jku:
I've done some testing with:
* ML-DSA in general -- offline and online signing works in https://github.com/jku/pqc-test-repo2
* yubikey signing: the new pkcs11 stack seems fine in tests in tuf-demo
* TKey support: looks ok in https://github.com/jku/pqc-test-repo2
* the signing key caching in ~/.local/share/tuf-on-ci-sign/signing-keys.ini

I'm ok with the state of things and will merge now. Futher review comments are welcome in new issues

PR #843, opened 2026-08-27
Describe what actually changed in an online delegation

### LINKED ISSUE #199: bug: wrong description of online role change
changing online expiry and signing periods resulted in this output from tuf-on-ci-delegate when signing:

```
 * Modified online delegations timestamp & snapshot
   * Signers: 1/1 of ['gcpkms']
     (was: 1/0 of []
```

That's not useful or even really correct

arpitjain099:
Fixes #199.

Two separate things are behind that output.

The `(was:` line is never closed. `_role_status_lines` has the same omission on its `(expiry period was ...` line, so that one is fixed here too.

The more useful half is why signers are being described at all. A delegation `Role` carries `x-tuf-on-ci-expiry-period` and `x-tuf-on-ci-signing-period` in its unrecognized fields next to the keyids and threshold, so changing a period alone makes `role != old_role` true. The block underneath only ever printed signers, which is why a period change came out as a signer change that had not happened.

So this prints the signer line only when the threshold or the signers actually differ, and prints a changed period on its own line. For the case in the issue:

```
 * Modified online delegations timestamp & snapshot
   * Expiry period: 10 days (was: 4 days)
```

I did not chase the `1/0 of []` in your paste. `_get_keys(known_good=True)` deliberately skips known-good keys that carry neither `x-tuf-on-ci-keyowner` nor `x-tuf-on-ci-online-uri`, for the repo-import case, and I could not tell from here whether your repository hits that path or something else. With this change the line is not printed for a period-only change anyway, but if the empty list shows up on a genuine signer change it is worth a separate look.

## Testing

Three cases in `signer/test/test_signer_repository.py`, driving `_delegation_status_lines` against a copy of `repo/test/test_repo1`: a period change, a threshold change, and no change at all. The threshold case asserts the `(was:` line ends in `)`, so it covers the paren rather than leaving it to review.

Against `main`, the first two fail:

```
FAILED test/test_signer_repository.py::TestDelegationStatusLines::test_expiry_period_change_is_described
FAILED test/test_signer_repository.py::TestDelegationStatusLines::test_threshold_change_is_described
E       AssertionError: False is not true :      (was: 1/1 of ['envvar']
```

One thing worth your eye: the fixture builds the repository with `SignerRepository.__new__` and sets `_dir`/`_prev_dir`, because producing delegation status lines needs nothing else and setting up a full signing event for this seemed disproportionate. Happy to build a real one instead if you would rather the test go through `__init__`.

The fixture also bumps root to v2 in both copies, since `_delegation_status_lines` only populates `old_delegations` when `root.version > 1` and everything reads as "New" at v1.

`ruff check signer`, `ruff format --check signer` and `pytest signer/test` are clean. `mypy signer` reports two pre-existing errors in `_user.py` and `_common.py`, which I did not touch and which also appear without this change.

### DIFF
--- a/signer/src/tuf_on_ci_sign/_signer_repository.py
+++ b/signer/src/tuf_on_ci_sign/_signer_repository.py
@@ -676,7 +676,7 @@ class SignerRepository(Repository):
             if signed.version != 1:
                 output.append(
                     f"   (expiry period was {old_expiry} days, "
-                    f"signing period was {old_signing} days"
+                    f"signing period was {old_signing} days)"
                 )
 
         return output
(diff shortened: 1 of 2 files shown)


jku (approved):
Thanks, that looks better.

Would probably be good to not yoink the test data from the other sub project... but it's fine, should be easy to cleanup if it becomes a problem

